PR #11, opened 2021-05-11
Add inlet / outlet assistance

### LINKED ISSUE #2: Verbose information when hovering over outlets
As per this forum thread:
https://discourse.flucoma.org/t/verbose-information-when-hovering-over-outlets/

When hovering over the outlets for Max objects, no information is given about the output, which is unlike most native Max objects.

weefuzzy:
* resolves #2
* see also flucoma/flucoma-core#12
* Adds inlet / outlet information for all objects: RT objects need their own individual strings & buf* objects / model objects have uniform labelling across objects 
* This also de-noises modelling objects by removing useless signal inlet and threading attributes / process() selectors

### DIFF
--- a/source/include/FluidMaxWrapper.hpp
+++ b/source/include/FluidMaxWrapper.hpp
@@ -59,11 +59,22 @@ class RealTime
 public:
   static void setup(t_class* c)
   {
-    class_dspinit(c);
-    class_addmethod(c, (method) callDSP, "dsp64", A_CANT, 0);
-    class_addattr(c, attribute_new("latency", USESYM(long), 0,
-                                   (method) getLatency<Wrapper>, nullptr));
-    CLASS_ATTR_LABEL(c, "latency", 0, "Latency");
+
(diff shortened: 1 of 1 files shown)


weefuzzy:
I don't think we've set things up to require reviews, so it's more just a matter of confirming in the first place that the changes are tolerable, and in the second place that catastrophic things don't happen. On that basis, if you're working in Max for anything, then it might be worth working off this to make sure that nothing breaks and that you're happy enough with the inlet / outlet text.  I *shouldn't* make any odds at all to SC, but because the Clients have been touched, there's always the Possibility of Surprise ™️ .

tremblap:
ok so one safe approach would be to compile sc on it, run the UTs, then compile max on them next time I spend time in Max so no shit hit the fan. I'll see if i can squeeze that in the next few days

tremblap:
It seems to be working at first glance. The (unused) label is helpful if a little disconcerting ;) I'll continue playing with them

PR #29, opened 2021-09-14
ci/automatic building

jamesb93:
This automatically builds the FluCoMa externals on a macos runner with github actions on each commit to main. This will help to automatically alleviate any issues where relatively _canon_ build processes with cmake are broken by rogue commits or edits to the `cmake` configuration. Could potentially just run on tagged commits or to check PR's.


jamesb93:
Further to this, we can also use this action to eventually build and publish nightly releases as well as stable releases with tagged commits.

weefuzzy:
Nice. What's a sensible way to test this? 

Longer term, it would indeed be nice to be able produce releases this way, although there's still a certain amount of human tidying up that has to be done (test-only externals have to be removed, for instance).

jamesb93:
To test that the automatic building works? Not sure. I can send proof of it working on my fork? I'm not sure how to run an action on a PR branch...

 Right now I flew a bit closer to the sun and broke it by trying to run max patches headlessly as part of the build process. I'll dial it back to the state in which it compiled the package successfully.

jamesb93:
You can see on my fork the action running.

https://github.com/jamesb93/flucoma-max/runs/3609550172?check_suite_focus=true

We could tune it to be more selective, doing certain things on tagged commits or perhaps only running on PRs. Anyway, there is no urgency in closing this off and I intended to leave it here dangling so we can come back around to this when there are less workshop-y things pressing :)

jamesb93:
Apologies for force pushing, I rebased a few of my branches which were intermingled in order to separate out some of the commits which had polluted. Basically, I just squashed all of the commits I made while trying to get this CI to work so that it keeps the history a bit neater.

jamesb93:
This now builds but I haven't tested building with `-DDOCS=ON`. That's the next stage.

jamesb93:
At the moment this works but does not work if I compile with -DDOCS=ON for various reasons (I think python environments and such are making `cmake` sad). My strategy around this is to add another build component that will make the max references on Linux and then compile the results at the output.

jamesb93:
This is now finished.

https://github.com/jamesb93/flucoma-max/releases

At the moment its using an ad-hoc branch to pull down core where BLAS is not used in Eigen (which is a problem across all CCE's actually). This should be solved in a coming PR for core which I will then reflect in the CI to use the tip of `dev` rather than my branch (`fix/blas-compilation-settings`)

weefuzzy (changes requested):
Question: what is `patchers/write-test.maxpat` for? Should it actually be in version control?

That notwithstanding, we should merge asap after the BLAS PR is in.

jamesb93:
> Question: what is `patchers/write-test.maxpat` for? Should it actually be in version control?
> 
> That notwithstanding, we should merge asap after the BLAS PR is in.

Oops. I was a bad boy, testing out if we could launch patchers headlessly for some kind of egregious Max-based testing. I'll `git rm` this.

jamesb93:
> Question: what is `patchers/write-test.maxpat` for? Should it actually be in version control?
> 
> That notwithstanding, we should merge asap after the BLAS PR is in.

I've removed it from vc, so once https://github.com/flucoma/flucoma-core/pull/67 is in we're good to go.

PR #36, opened 2021-09-20
Enable naming / sharing for all RT queryable model objects

weefuzzy:
Supports this PR: https://github.com/flucoma/flucoma-core/pull/35

Besides a bug fix in the wrapper to do with making `attrui` refresh when another instance's params change, this consists just of updating the libmanipulation.cpp 

I do have locally some scraggly additions to some help files that I used to test, but don't want to interfere with the other branches doing updates to help files.


tremblap:
working - merged into release candidate branch - will close when merging in main - still raises questions for attributes saring between objects but we can tweak the interface I think.

jamesb93 (approved):

PR #38, opened 2021-09-20
fluid.plotter

jamesb93:
This is a piece of JSUI code and an accompanying help file which can be used for plotting points in two-dimensional space. It also has an in built distance calculation for getting the nearest point to the mouse out, rather than having to loop around with the raw coordinates and KDTree.

### DIFF
--- a/help/fluid.plotter.maxhelp
+++ b/help/fluid.plotter.maxhelp
@@ -1,0 +1,991 @@
+{
+	"patcher" : 	{
+		"fileversion" : 1,
+		"appversion" : 		{
+			"major" : 8,
+			"minor" : 2,
+			"revision" : 0,
+			"architecture" : "x64",
+			"modernui" : 1
+		}
+,
+		"classnamespace" : "box",
+		"rect" : [ 35.0, 88.0, 891.0, 777.0 ],
+		"bglocked" : 0,
+		"openinpresentation" : 0,
+		"default_fontsize" : 12.0,
+		"default_fontface" : 0,
+		"default_fontname" : "Arial",
+		"gridonopen" : 1,
+		"gridsize" : [ 15.0, 15.0 ],
+		"gridsnaponopen" : 1,
+		"objectsnaponopen" : 1,
+		"statusbarvisible" : 2,
+		"toolbarvisible" : 1,
+		"lefttoolbarpinned" : 0,
+		"toptoolbarpinned" : 0,
+		"righttoolbarpinned" : 0,
+		"bottomtoolbarpinned" : 0,
+		"toolbars_unpinned_last_save" : 0,
+		"tallnewobj" : 0,
+		"boxanimatetime" : 200,
+		"enablehscroll" : 1,
+		"enablevscroll" : 1,
+		"devicewidth" : 0.0,
+		"description" : "",
+		"digest" : "",
+		"tags" : "",
+		"style" : "",
+		"subpatcher_template" : "",
+		"assistshowspatchername" : 0,
+		"boxes" : [ 			{
+				"box" : 				{
+					"id" : "obj-107",
+					"maxclass" : "comment",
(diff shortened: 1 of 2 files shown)


jamesb93:
As per the discussion this morning, I have updated the interface to be more "fluid" with just the output of a dumped dataset for drawing the points, and a dumped labelset for colouring them. I have made examples covering how this works in the associated help file.

Any further feedback very welcome 🦖

jamesb93:
https://github.com/flucoma/flucoma-max/pull/38/commits/000a10f85e312d3e63be29680c6a4302f630964f adds the possibility to scale each dimension independently. This makes the plotter more flexible in visualising data that is not within a 0 -> 1 range in both dimensions

jamesb93:
>     * it would be good that the first button reset the plotter so no labels are still in there
Done :) Naughtily wasn't clearing internal colorMap on reset.

>     * maybe the background could be greyer than full white
Went for 0.9, 0.9, 0.9 in RGB space

>     * a bug in the hightlight size boost: it is not centred so feels weird
This was an overall sloppy bit of coding where I was drawing from the top left of the position. It now takes half of the square/circle size and offsets the position so it is centred.
 
>     * we also talked about adding a frame to those selected - is that gone?
As per office discussion it is flavour that is perhaps not needed. From what my noob eyes can see, there isn't a simple way to determine an outline for a shape when you draw it. Could be faked with an additional box that is slightly bigger and has a black fill.

jamesb93:
The messages `setdict` `setcategories` should be made **much** simpler so that a load of Max routing doesn't have to be done in order to get the dictionary information...

weefuzzy:
Unfortunately the bottom option isn't a possibility at the moment: I don't trust the polyglot stuff in Max enough to give us a way of talking to a dataset from JS. 

The middle option feels the most idiomatic to me, as it means that `fluid.plotter` responds to the message `dictionary <blah>` in the same way as other `dict`-aware objects. For categories, I think I would use a another `dictionary` message into a second inlet. 

I don't especially like the top one because we have non-idiomatic interface (`set<stuff>`).

jamesb93:
My only gripe with making inlets is it can make patching unwieldy, especially if you want a big plotter. It does simplify things a little though in terms of what goes where and what messages mean and allows us to just pass dicts in willy nilly.

weefuzzy:
Swings and roundabouts, I think. Certainly a two-inlet native dictionary interface has familiarity on its side 😄

jamesb93:
Is there a way we can guard against a person bungling the inputs? Maybe check that there is in fact the right looking data crudely? Maybe just belch if the dimensions are wrong (which would catch a lot of this scenario).

weefuzzy:
In what scenarios? When being passed a `dict`? Yes – we can check that the keys are as we expect etc.

jamesb93:
> In what scenarios? When being passed a `dict`? Yes – we can check that the keys are as we expect etc.

Yep, we can check the shape of the data (1 col/ 2 col). I think that is the only thing right now which differentiates `labelsets~` and `datasets~` when you dump them.

weefuzzy:
We'll expect the dataset's entries in `data` to be arrays of floats,  whereas the labelset can be strings.

jamesb93:
> We'll expect the dataset's entries in `data` to be arrays of floats, whereas the labelset can be strings.

Smashing - i'll put this together later today

jamesb93:
I've made it "backwards compatible" in that one can pass dictionaries to the appropriate inlet, or you can provide the setdict and setcategories message. ATM the two messages are not shown in the help file, but they are there if someone prefers that patching style.

I also added an advanced tab to the help file to show how the plotter can be constructed programatically _without_ a dataset~ at hand. 

This might constitute the final iteration I would think so if you get a chance to have a look let me know @weefuzzy @tedmoore @tremblap

PR #39, opened 2021-09-22
make fluid.bufview not complain

jamesb93:
If you create a fluid.bufview. jsui object and resize it without setting a buffer it complains endlessly in the max console. I have guarded against trying to downsample on resize if the buffer hasn't been set, i.e the user has just made the object.


jamesb93:
Apologies, I fixed this on a branch with other changes. Will just pull it directly into main.

PR #69, opened 2021-10-29
reconfigure fluid.plotter interface

jamesb93:
This adds several changes to the fluid.plotter object. Fundamentally it changes how the data of the fluid.dataset~ dump is dealt with.

1) No longer are the points in the plot an `array`. They are now a `JSON Object`. This has the consequence of each point being a key-value pair with information about its position, size and colour. This means that points are now unique. This has consequences on the `addpoint` message below.

2) `addpoint` now has mirrored functionality with `fluid.datasets~` method. If the point already exists, and you try to use the advanced interface to manually add a point with the same identifier it will throw an error.

3) To make the changes for `addpoint` feel like one is losing something, `setpoint` now exists for changing the properties of individual points after the fact.

### DIFF
--- a/jsui/fluid.plotter.js
+++ b/jsui/fluid.plotter.js
@@ -124,7 +124,7 @@ function setdict(name) {
 		fail = true;
 	}
 	if (dataDict.get('cols') != 2) {
-		error('fluid.dataset~ should be exactly two dimensions.', '\n')
+		error('fluid.dataset~ should be exactly two dimensions', '\n')
 		fail = true;
 	}
 	if (!fail) {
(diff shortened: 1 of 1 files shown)


jamesb93:
I'm also thinking that one should be able to alter the colour of points after the fact in the same way their position can be altered with `setpoint`. I would prefer not to be back in `jit.lcd` land with 7 argument messages and would rather have separate messages for setting the position with identifer and colour.

ATM there is `pointcolour` but maybe there can be `setpointcolour` too?

What y'all think?

jamesb93:
As a note of why these changes were reverted, we had a meeting on 3 November discussing the interface across Max, SC and PD.

Ted was concerned that having bundled up methods for ranges (xrange, yrange and range) would make building interfaces around the plotter difficult, because you would always have to retrieve the state of multiple GUI objects to construct the correct messages. For Max this isn't really an issue because state is handled entirely differently.

As such, the interface will diverge in this aspect with PD and Max having `xrange`, `yrange` and `range` again and SC having `xmin`, `xmax`, `ymin` and `ymax`.

PR #75, opened 2021-11-08
Add relative path file writing for data objects

### LINKED ISSUE #20: enable relative paths for data objects' `write` message
Whenever I call the `write` method, for example with the mlpregressor~, I get the error:

`fluid.mlpregressor~: Could not open file for writing`

I have checked the system security preferences in order to rule out some permissions, but I have given full disk access to Max. What would be a way in to debugging out any stupid decisions I may have made patching?

weefuzzy:
closes #20 

In principle consistent with other Max objects (which aren't wholly consistent in edge cases). I've tried to do what `[dict]` does

* Absolute paths should pass through unchanged 
* A path-less file name passed to `write` in a *saved* patch will use the same folder as the patch 
* A path-less file name passed to `write` in a *unsaved* patch will use the same folder from which the last patch was opened 
* If there was no last-opened-patch (e..g unsaved patch after Max has just started, no re-opened windows), like `dict` (and `jit.matrix`) files end up in `/Applications` (!!) [this isn't me, this is what the Max API returns as a candidate path 🤷‍♂️ ] 

Needs verifying on Max 7 and Windows

### DIFF
--- a/source/include/FluidMaxWrapper.hpp
+++ b/source/include/FluidMaxWrapper.hpp
@@ -1738,7 +1738,36 @@ private:
       path_toabsolutesystempath(path, filename, fullpath);
     }
     else
-      strcpy(fullpath, s->s_name);
+    {
+      char  thispath[MAX_PATH_CHARS];
+      char  infolder[MAX_PATH_CHARS];
+      path_splitnames(s->s_name, infolder, thispath);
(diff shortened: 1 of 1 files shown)


jamesb93:
This set of behaviours seems consistent to me. I just tested it and it works too which is nice 😂

jamesb93 (approved):

PR #84, opened 2021-11-23
adds to objectlist

tmhglnd:
This will show the plotter and waveform in the autocomplete objectlist when typing a new object. Just a small contribution, but hopefully useful :).


jamesb93:
This is incredibly useful. Thanks so much!

PR #90, opened 2021-12-09
Remove spaces from installation path name

jamesb93:
Previously when running something like `ninja install` the package folder in `release-packaging` would contain spaces. This removes those spaces.﻿

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,7 +128,7 @@ endforeach ()
 #install bits. Max packages bundle Windows and Mac builds together
 set(CMAKE_INSTALL_PREFIX "${CMAKE_CURRENT_SOURCE_DIR}/release-packaging")
 set(MAX_INSTALL_PREFIX "." CACHE PATH "Prefix for assembling max package")
-set(FLUID_PACKAGE_NAME "Fluid Corpus Manipulation" CACHE STRING "Name for published package")
+set(FLUID_PACKAGE_NAME "FluidCorpusManipulation" CACHE STRING "Name for published package")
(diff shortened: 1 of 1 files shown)


weefuzzy:
My review might have said: we should also update `package-info.json.in` to be accurate (which needed to happen anyway, although some details are pending a discussion about how to better set version data).

tremblap:
I compiled and it worked so I merged like a punk. Feel free to fix right on dev like a punk :) The good news is that the other 2 CCE are in line with this so no changes needed there :)

tremblap:
oh look Windows build it failing : https://github.com/flucoma/flucoma-max/actions/runs/1566980850

tremblap:
So I need to understand the PR dance a bit better - one is expected to test on all OSes before approval?

weefuzzy:
In this case the failing build is because the CI itself needs updating to reflect this change, so testing locally wouldn't have made much difference. 

In general, one would assume that the first responsibility for testing lies with the person submitting a PR. 

The state we would wish to move towards is that this is verifiable through tests where applicable, so that verification happens through CI as well. This means that good bug fix , e.g., would include a test that reproduces an issue before fix and passes post-fix. 

At the moment, because we are test-poor, it's necessarily more of a judgement call. Something complex or scary might well warrant some building and testing by reviewers. Other things, not so much.

jamesb93:
I am addressing at least putting in a cursory check to see if CI breaks on PRs which is an indication if you may have either broken CI or the ability to build on one of the combinations:

https://github.com/flucoma/flucoma-max/pull/95

weefuzzy:
So I see 😄

PR #91, opened 2021-12-09
[Review Changes] BufSelect Help File

jamesb93:
Adds a small update to the BufSelect help file.﻿


jamesb93:
1. Neaten up inside of fill draw
2. Irregular shape for buffers (3x4 and 2x3 for in/out)
3. Reflect shape in the `matrixctrl` output
4. Remove word cherry-pick in favour of something less vernacular
5. Don't use symmetrical indices/channels arguments
6. Stats picker for another tab / centroid

PR #92, opened 2021-12-09
[Review Changes] KDTree Help File

jamesb93:
Small cosmetic upgrades to the KDTree Patch﻿


jamesb93:
I think that this (and all help files where possible) should link to the video references and other more substantial patches showing the use cases. In this case KDTree is very relevant to the sound plotting patch, with both 2 and more dimensions.

jamesb93:
1. Radius and numneighbours should be second tab
2. Mouseover instructions
3. KDTree first tab should be 1 number
4. "more neighbours"
5. The initial example is way too abstract
6. A musical example at the start
7. Example sandwich? Many examples to show the spectrum of musical possibilities

PR #94, opened 2021-12-10
[Review Changes] Chroma Help File

jamesb93:
Updates to the fluid.chroma~ help file.


jamesb93:
1. Remove control of smoothing
2. One liner for pitch class
3. Cute musical example with `ioscbank~`, `comb~`, fffb~`. Should come in the first tab.
4. Observing the multislider (remove these kind of redundancies)
5. Jargon in reference (class, histogram, pitch class)
6. Remove fftsettings from first tab and make a more precise fftsetting in later tabs
7. `zl.change` in second tab
8. You can think of it is as... typo
9. It doesn't have to be 12 divisions

PR #98, opened 2021-12-13
[Review Changes] BufChroma Help File

jamesb93:
Improvements to fluid.bufchroma~ help file.﻿


jamesb93:
1. Make fluid.help.js larger so text can be seen
2. Musical example? Offline version of real-time chroma example
3. Each channel is reserved for one chroma
4. Confusing sentence on the first object argument
5. Everything is hanging out on the first tab of the help file. Start similarly austere and pack into two tabs.
6. First tab piano, second tab oboe

PR #102, opened 2021-12-16
Load media files from package abstraction

jamesb93:
This adds a new abstraction that can be used in a bpatcher for helping users load sounds from the media folder into buffers.

It also constructs an internal lookup table so you can set a default, ala @file filename.wav.


jamesb93:
@weefuzzy if you are happy with this (and also certain that jstrigger can't save us from the js sins) then can you approve this one and one of us will merge?

PR #103, opened 2022-01-06
[Docs] Sines Help File

jamesb93:
A few cosmetic upgrades to make the visual aesthetic normalised with our own style as well a the broader max conventions of docs.﻿


tremblap (commented):
maybe we should put in the flucoma player a default sound that works (like the piano)

otherwise it is indeed a lot cleaner and easy to read!

PR #104, opened 2022-01-06
[Docs] Spectralshape Help File

jamesb93:
- Overhaul structure of help file
- Remove redundant tab explaining all the different outputs (merged into the nice bpatchered tab)
- Simplify the final tab
- Reorder to make musical example paramount﻿


tremblap (changes requested):
daring - first tab with log centroid (I love that example) - I think you should put logarithmic in front of the explanation (The logarithmic spectral centroid and its spread are used to drive ...)

I wonder if my extensive tutorial (2nd and 3rd tab) shouldn't go on the website with the code copy/pastable? I love it there but is it consistent with the other objects/approaches?

PR #105, opened 2022-01-07
[Docs] Audiotransport Help File

jamesb93:


tremblap (commented):
clearer indeed. I'm not sure the use of thru floats my boat (it is ugly) but it does make it clearer to read (less ugly) -  the impact of fft size is probably the best smearing example we can provide in the STFT overview I just read.

general question: now that we have 'online references' I wonder if we shouldn't add a button to the local reference too (to launch the right page in the reference window)

jamesb93:
> clearer indeed. I'm not sure the use of thru floats my boat (it is ugly) but it does make it clearer to read (less ugly)

Never been a fan either

> general question: now that we have 'online references' I wonder if we shouldn't add a button to the local reference too (to launch the right page in the reference window)

I would opt not to do this because the `?` tab already serves this purpose and reduces the amount of UI layout business we need to manage. I also think its not that conventional to launch the reference from the patch.

jamesb93:
> > clearer indeed. I'm not sure the use of thru floats my boat (it is ugly) but it does make it clearer to read (less ugly)
> 
> Never been a fan either
> 
> > general question: now that we have 'online references' I wonder if we shouldn't add a button to the local reference too (to launch the right page in the reference window)
> 
> I would opt not to do this because the `?` tab already serves this purpose and reduces the amount of UI layout business we need to manage. I also think its not that conventional to launch the reference from the patch.

PR #110, opened 2022-01-20
[Docs] Grid Help File

jamesb93:
Update the fluid.grid~ helpfile


jamesb93:
I've had a crack at reworking the grid example. Struggling to find a non-trivial musical example here without having to make a bunch of datasets, explain what is going on and then hope they find the connection to the sound that it makes... maybe it is clear enough its usefulness from the three visual examples?

tremblap (changes requested):
I was surprised at first with the shapes as it stays in the visual realm (instead of music) but this is at least very graphic and much clearer. The only questions are

1. if we are going to use these demo .json datasets across all CCEs should we not put them in the core
2. the extent int should be capped at 0 at the bottom

jamesb93:
> I was surprised at first with the shapes as it stays in the visual realm (instead of music) but this is at least very graphic and much clearer. 

I pondered having a sample-selector type interface but I'm not sure if its too much to bake in to the example? That would be my go to audio example to code up but I am just wary of making it feel like a big MuBu style help patcher where its very hard to dissect what the relationship between all the bits are.

The only questions are
> 
> 1. if we are going to use these demo .json datasets across all CCEs should we not put them in the core

That's definitely an idea. I generated them from Python, which itself has a module for generating all sorts of data set archetypes. Maybe @weefuzzy would suggest to translate those generators for the CCEs instead to make them more modular and re-usable? Takes time to code though.

> 2. the extent int should be capped at 0 at the bottom

Good catch. Maybe an upper limit too as it becomes chunky to process otherwise?

jamesb93:
This is now dependent on https://github.com/flucoma/flucoma-core/pull/88 and #126.

PR #116, opened 2022-01-25
[Enhance] Update fluid.waveform~

jamesb93:
Updates fluid.waveform~ to have parity with the SuperCollider waveform.

`addlayer line` is now `addlayer featurebuf`.

`addlayer image` is now `addlayer imagebuf`

`addlayer wave` is now `addlayer audiobuf`

All calls are backwards compatible too.

### DIFF
--- a/help/fluid.waveform~.maxhelp
+++ b/help/fluid.waveform~.maxhelp
@@ -3,14 +3,14 @@
 		"fileversion" : 1,
 		"appversion" : 		{
 			"major" : 8,
-			"minor" : 1,
-			"revision" : 11,
+			"minor" : 2,
+			"revision" : 2,
 			"architecture" : "x64",
 			"modernui" : 1
 		}
 ,
 		"classnamespace" : "box",
-		"rect" : [ 34.0, 79.0, 1394.0, 956.0 ],
+		"rect" : [ 34.0, 87.0, 988.0, 726.0 ],
 		"bglocked" : 0,
 		"openinpresentation" : 0,
 		"default_fontsize" : 12.0,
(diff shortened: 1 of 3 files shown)


tremblap (changes requested):
great, thanks. A few things to fix:

the interface we spoke of was:
audioBuffer
imageBuffer 
featureBuffer 
indicesBuffer 

so for now I think we should
- change *buf to *buffer
- change markers to indiceBuffer (no camel for max I know)
- remove the generic 'buffer' message



I'm not sure I'd leave the backward compatibility although I understand we are exploring interface here. We'll need to break something at some point but I'm happy to wait for this last point of breaking backward compatibility.

jamesb93:
> great, thanks. A few things to fix:
> 
> the interface we spoke of was: audioBuffer imageBuffer featureBuffer indicesBuffer
> 
> so for now I think we should
> 
> * change *buf to *buffer
> * change markers to indiceBuffer (no camel for max I know)
> * remove the generic 'buffer' message
> 
> I'm not sure I'd leave the backward compatibility although I understand we are exploring interface here. We'll need to break something at some point but I'm happy to wait for this last point of breaking backward compatibility.

Okay, I'll need to change these names as well as dig around the markers API too 😱

jamesb93:
Paging @tremblap. Those changes are now in place.

tremblap:
oh and now the pedantic english language dude in me says: we have indicesbuffer (plural) and featurebuffer (singular)

I presume at this point it is relatively easy to change but my anxiety level is high: which should it be

audioBuffer
imageBuffer
(so far so goo)
featuresbuffer vs featurebuffer
indicesbuffer vs indicebuffer

sorry for finding that out now. interface consistency, your worse nightmare. btw @tedmoore that implies a SC change either way we go! I won't vote though apart from having one of each :)

jamesb93:
> 

It should be singular! Good catch. I'll make this so.

jamesb93:
I've also fixed the typos @tremblap

tedmoore:
The plural of `indices` is just `index`. I think `indexBuffer` is confusing because it implies that it will (or should?) have just one index, but it will almost always have many. Same with features, even if it is just one channel, all the frames are features, aka analysis results.

My vote would be that they all be plural, although I don't feel strongly about it either way. 

We should avoid `indiceBuffer` though because indice is an incorrect spelling.

jamesb93:
> The plural of `indices` is just `index`. 

Woops! 😂

How about:

indicesBuffer
featuresBuffer
audioBuffer
imageBuffer

That feels consistent enough to me.

tedmoore:
i like it

Issue #111, opened 2022-01-20
Fluid.bufchroma help: a few errors


tremblap:
Love the patch. A few problems, from our emails

1. the text from the doc is cropped (see picture below)
2. "an histogram" should be "a histogram" and "class" should be "object" as I was told by @jamesb93 
3. there are bufcompose errors every time it loads a file
4. in the 'more details' we need (like the fluid.chroma~) a way to adapt to varying number of chroma bin. the code is almost there. it will be good to show users how to dismiss the spare channels.

![Screenshot 2022-01-20 at 15 39 46](https://user-images.githubusercontent.com/10000381/150415844-b6a85edb-47b7-456d-bbc3-5e3d243060d6.jpg)

jamesb93:
> 1. the text from the doc is cropped (see picture below)

That is solved by changing the layout of the patch now in the linked PR.

> 2. "an histogram" should be "a histogram" and "class" should be "object" as I was told by @jamesb93

That is a docs issue not a Max issue so will be solved there.

> 3. there are bufcompose errors every time it loads a file

bug squashed


> 4. in the 'more details' we need (like the fluid.chroma~) a way to adapt to varying number of chroma bin. the code is almost there. it will be good to show users how to dismiss the spare channels.

AFAICT it will adapt in more details. The below screenshot shows the part of the patch that calculates bin frequencies.

![image](https://user-images.githubusercontent.com/13281862/150977578-4b1528d6-3449-455a-a29d-d15e99fc4004.png)

tremblap:
good for 1-2-3

#4: maybe the 'more details' can be streamlined to not do audio (like what I proposed in the Pd version) as it is 'more advanced' and there are still complicated points to explain (why the empty channels) ?

tremblap:
also, the current patch doesn't adapt - it should not adapt to the number of channels (as it will always be 19) but to the numchroma, which will need to be passed differently than it is right now. But again if we go for my proposal it is a moot point but this error shows we need to explain the 19 channels of buffer one will get, of which numchroma only are non-zero/used/computed

jamesb93:
> also, the current patch doesn't adapt - it should not adapt to the number of channels (as it will always be 19) but to the numchroma, which will need to be passed differently than it is right now. But again if we go for my proposal it is a moot point but this error shows we need to explain the 19 channels of buffer one will get, of which numchroma only are non-zero/used/computed

[...]

PR #117, opened 2022-01-25
[Docs] Fix Layout issues in BufChroma help

jamesb93:
Fixes layout issues described in:

https://github.com/flucoma/flucoma-max/issues/111


tremblap (approved):
rock on - I'll push the doc change right on dev now.

PR #118, opened 2022-01-25
[Docs] Loudness Help File

jamesb93:
Updates the loudness help file.﻿

### DIFF
--- a/help/fluid.loudness~.maxhelp
+++ b/help/fluid.loudness~.maxhelp
@@ -10,7 +10,7 @@
 		}
 ,
 		"classnamespace" : "box",
-		"rect" : [ 1052.0, 442.0, 826.0, 615.0 ],
+		"rect" : [ 1052.0, 442.0, 910.0, 743.0 ],
 		"bglocked" : 0,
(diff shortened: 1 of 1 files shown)


jamesb93:
@tremblap for the same reason I requested review on the learn reference for loudness I believe you are the resident expert on it :) Given our recent sync I would usually reserve learn for Ted and not double your reviews but I just can't resist picking your brain on it.

tremblap (changes requested):
example tab 1: to start with sound is super cool: proposed change: pink~ and *~2 so it sounds less aggressive and toppy.

example tab 2: 2 proposed mods: instead of this source, let's use the fluid one starting in oscilator mode as it allows people to explore more without repatching.
also the sentence that tells the ITU stuff needs changing because these values are true only at 44100 Hz. what about:

"These windowsize and hopsize settings are what the ITU BS 1770 specification defines for metering loudness of a signal. They define it at 400ms window updated every 100ms."

workflow wise: learn are for both @tedmoore and I, as far as I understood, but if you guys are happy and confident then let's do it like this so I recover some bandwidth.

PR #122, opened 2022-01-27
[Docs] BufLoudness Help

jamesb93:
Reworks the bufloudness help file﻿


tremblap:
very nice visualisation.

in the 2nd tab, is great, but maybe you should do bufstats of both loud and peak and get the max of true peak - that would be slick :)

i'm surprised to see a multichannel tab - didn't we propose to have a generic one for all bufprocesses (like the blocking attribute)

otherwise super slick!

jamesb93:
Good idea re: reusing help files to show multichannel-ness. I'll strip that out and add your suggested max true peak :)

PR #126, opened 2022-02-01
[Enhance] Retrieve toy DataSets

jamesb93:
Adjusts paths in the CMake configuration for where `AudioFiles` and the new `DataSets` (see https://github.com/flucoma/flucoma-core/pull/88) are retrieved from.﻿


jamesb93:
BTW I've tested this with the `install` target which works, but for working on the repo itself (as I think you do too) it doesn't seem to copy anything to the misc folder from core. Did I miss something as surely the audio files follow the same pattern.

jamesb93:
This works now and is just waiting on SC.

CMakeLists.txt
```diff
@@ -136,8 +136,10 @@ install(DIRECTORY ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/
         FILES_MATCHING REGEX "\\.(mxo|mxe|mxe64)")
 
 # install(CODE [[ message(STATUS "${flucoma-core_SOURCE_DIR}") ]])
-install(DIRECTORY "${flucoma-core_SOURCE_DIR}/AudioFiles/" 
+install(DIRECTORY "${flucoma-core_SOURCE_DIR}/Resources/AudioFiles/" 
         DESTINATION "${MAX_PACKAGE_ROOT}/media")
+install(DIRECTORY "${flucoma-core_SOURCE_DIR}/Resources/DataSets/"
```
weefuzzy, line 141: Copy  to the top level `misc` first and then let the install rule for all the top level folders take care of installing in the final destination. Then it will work for both dev and deploy.

weefuzzy: but we'd then need to .gitignore the json files in misc 🙄

jamesb93: AHHH

jamesb93: Thanks

PR #127, opened 2022-02-01
[Docs] UMAP Help File

jamesb93:
Updates the fluid.umap~ help file.﻿


tremblap (changes requested):
I love this very much. 

I get this error when I press the #1 of the first pane
dict.view: doesn't understand "pointsizescale"

also, using 'you' in the description is not super. 'one' is also bad. so maybe a passive voice?

2nd tab: when we say experiement with, we should tell people to re-press fit transform, no?

3rd tab: great example - we should credit where it is from though as it is a classic

PR #128, opened 2022-02-04
[Docs] BufScale Help File

jamesb93:
Updates the BufScale help file. Not a particularly sweeping set of changes more just normalising the interface with other updated help patchers.


tremblap (changes requested):
hard to be clearer :)

the layout is ugly at the top (a bit white space under the description) otherwise it is very clear. I notice that you removed the multichannel example, which would be good to keep in this case as it does channel-wise normalsiation (and not buffer wise like native buffer objects)

PR #130, opened 2022-02-08
[Docs] DataSetQuery Help File

jamesb93:
Updates the fluid.datasetquery~ help file.

I went with the approach of separating out the ideas as much as possible as it was all stacked on top of each other in the original verison. @tremblap would you be able to have a glance and tell me what you think?

### DIFF
--- a/help/fluid.datasetquery~.maxhelp
+++ b/help/fluid.datasetquery~.maxhelp
@@ -3,8 +3,8 @@
 		"fileversion" : 1,
 		"appversion" : 		{
 			"major" : 8,
-			"minor" : 2,
-			"revision" : 2,
+			"minor" : 3,
+			"revision" : 0,
 			"architecture" : "x64",
 			"modernui" : 1
 		}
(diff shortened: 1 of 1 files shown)


tremblap (changes requested):
very good structure
tab 1, the compositing in one message, I would add a different output to the query just to show it works, and verbose it (compositing the query as one message, this time asking for the 2nd column) that way we see the change of output with the same query and we see something moving :)
tab2, I would change the and condition to <= 10 to remove more information - and the or to <=5000. in the first case it excludes 1 more item, and in the latter in includes it all. that forces the user to try to understand the result they get more
tab3 I would remove the and/or option to decluter and keep the take home message only (range) and would do addrange 1-2 as it shows a serious effect of cutting both sides (and the range. I would probably put 2 examples (addrange 2-2 and addrange 1-2) for contrast
tab4- a sentence on the same ID being associated in the query should be there, probably in the blue box (and a typo in the pink box: satisfy is misspelt)

PR #133, opened 2022-02-10
[Docs] Normalize Help File

jamesb93:
Updates the normalize help file.

@tremblap do you mind taking a look at this?


tremblap (changes requested):
this is great again!
one typo in tab3 (learnt, not learn, minima ...)
tab3 could have a fittransform to observe the full results but that is a detail and might be clearer as it is, what do you think?

PR #137, opened 2022-02-16
[Docs] KMeans Help File

jamesb93:
This updates the KMeans help file.

I chose to remove the final tab in favour of moving that level of detail to an example/tutorial that will come later.


tremblap (changes requested):
tab 1 2 4 are perfect - I love those toy dataset they make it so clear!
tab 3 is a bit confusing still. it would be good to maybe space them out a bit more to at least get one in the middle, and not to use fitpredict but just predict and tell the user that we do not get the algo to try to fit anything, it just takes our barycentres are truth

PR #139, opened 2022-02-17
[Docs] BufSines Help File

jamesb93:
This is mostly a buffer translation of the sines help file.﻿


jamesb93:
> this has been merged but the changes were not actioned?

I'm in the middle of filing a separate issue now to loop back around for it. Those parameters are nightmarish. I don't understand how they all interact but we need to clear all these help file PRs in preparation for #140.

jamesb93:
https://github.com/flucoma/flucoma-max/issues/165

tremblap:
there are lot left to think about for #140 so let's see its horizon - but indeed the real-time version of sines will be better.

PR #141, opened 2022-02-18
[Docs] BufSelectEvery Help File

jamesb93:
Updates the bufselectevery~ help file.


tremblap (changes requested):
wow sneaky use of frombuffer to visualise!

the only thing we said we would do is to not make the first tab input simetrical so it is clearer what is the hop in each dimension. so an input grid of 5 x 9 would fit and we could use a hop of 2 and 3 and that way the magic is clearer, no?

tremblap:
I also get this error: 
fluid.bufselectevery~: Source Buffer Not Found or Invalid

PR #143, opened 2022-02-21
[Docs] BufOnsetSlice Help File

jamesb93:
- update onsetslice (rt) help file
- start work unpeeling bufonsetslice

### DIFF
--- a/help/fluid.bufonsetslice~.maxhelp
+++ b/help/fluid.bufonsetslice~.maxhelp
@@ -10,7 +10,7 @@
 		}
 ,
 		"classnamespace" : "box",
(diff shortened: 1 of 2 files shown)


tremblap (changes requested):
tab1: because we visualise the slice it would be good to add the threshold parameter to people can wiggle right away and see how well it works.
tab2 is great - added visualisation makes it pop! - maybe we could add a zoom so we can see the more dense ones in more detail

tremblap:
tab2: we could add the parameter that are changed (attrui would do) so people can see and it is less abstracted

PR #144, opened 2022-02-23
remove fluid.waveform copy and download instead

weefuzzy:
Rather than the hilarity of having a separate copy of `fluid.waveform~` in the max repo, this downloads the latest build at configure time, and installs the components during install. 

It doesn't try to doing anything clever to the development folder.


jamesb93 (approved):
Looks good. We manage our own waveform business in development and packaging soaks up the latest release. Sounds good.

PR #146, opened 2022-02-24
[Docs] NoveltySlice Help Files

jamesb93:
- update noveltyslice
- shell out bufnovelty


tremblap:
The parameters are indeed a bit complicated. I usually play with it the way @g-roma recommended:
1. pick a feature, 
2. put the threshold to 0
3. play with the kernel size and filter size until it almost make sense in term of what I expect novelty 'is'
4. tweak threshold and minslicelen to tame the last undesired slices.

does that help?

PR #155, opened 2022-03-10
[Docs] BufCompose Help File

jamesb93:
Updates the bufcompose help file. Not much changed apart from aesthetics and consistency.


tremblap (changes requested):
very clear first tab. The main page is truncating the text (we should see it in full in tab 1 and not at all in tab2+ I think)
Screenshot 2022-03-23 at 16 21 18
:

tab2: destgain could be $1 and float so people can experiment without editing the patch - also we should put a clear on the destination if that becomes more playful and explorable (since it will pile up!)

tab 3: is there a way for the composite message to look less crammed? maybe a few annotated messages so it is simpler to read?

tab 4 is good - channels will be repeated , should we add starting from 0?

all other more advanced examples have disappeared. either we should point at them in the example folder (they were all very useful to showcase the flexibility of it) or added back (but I understand the desire of streamlining)

jamesb93:
> very clear first tab. The main page is truncating the text (we should see it in full in tab 1 and not at all in tab2+ I think) Screenshot 2022-03-23 at 16 21 18 :
> 

https://github.com/flucoma/flucoma-docs/pull/95 will make it shorter and fit into the space I've given it.

> tab2: destgain could be $1 and float so people can experiment without editing the patch - also we should put a clear on the destination if that becomes more playful and explorable (since it will pile up!)

good idea

> 
> tab 3: is there a way for the composite message to look less crammed? maybe a few annotated messages so it is simpler to read?

done

> tab 4 is good - channels will be repeated , should we add starting from 0?

Not sure I understand.

> all other more advanced examples have disappeared. either we should point at them in the example folder (they were all very useful to showcase the flexibility of it) or added back (but I understand the desire of streamlining)

They will find a new home in the FluCoMa tutorial system and be linked in a new tab.

tremblap:
> Not sure I understand.

if one has a stereo input, ask for numChans 2 and startchan 1, one will get the stereo inversed. this was one of the advanced examples that have been removed - they were quite involved so the help might not be the best place for them, but we need to explain that channels are not 'copied' but start back to 0 (round robin? modulo?)

> They will find a new home in the FluCoMa tutorial system and be linked in a new tab.

I saw the new issue. indeed a sort of vignette for advanced bufcomposing is probably the best way forward to declutter the help file and not get a [groove~] one

tremblap:
I know, but this sentence is misleading of the underlying behaviour. let's find an alternative elegant way of saying what it does

jamesb93:
> I know, but this sentence is misleading of the underlying behaviour. let's find an alternative elegant way of saying what it does

image

?

tremblap:
that will need @weefuzzy wordmagic here I think :)

PR #156, opened 2022-03-11
[Docs] BufSpectralShape Help File

jamesb93:
A small update to the bufspectralshape help file.

### DIFF
--- a/help/fluid.bufspectralshape~.maxhelp
+++ b/help/fluid.bufspectralshape~.maxhelp
@@ -3,14 +3,14 @@
 		"fileversion" : 1,
 		"appversion" : 		{
 			"major" : 8,
-			"minor" : 1,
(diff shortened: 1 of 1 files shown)


tremblap (changes requested):
very streamlined and a good first tab. 

- the patcher to 'retrieve' should be commented to help people understand the multichannel layout
- an explanation of the multichannel layout is missing - it is far from trivial as we saw in workshop for people to understand that 1 audio channel generates 7 descriptor time series

I think we need a tab to explain the channel structure of time series - especially for multichannel inputs - we talked before to create a common tab for all buf descriptor objects (all the new 'curved' ones will need that too)

I definitely miss the tutorial on what spectralshapes are and the log example. I know they are in the RT version but maybe we should invite people there with a sentence and a link?

PR #158, opened 2022-03-16
[Feature] Plotter drag to zoom

### LINKED ISSUE #148: datasetquery crashes max
load the included dataset in the included patch

pressing 55 (which has an entry) works

pressing 54 (which has no entry) crashes
[Archive.zip](https://github.com/flucoma/flucoma-max/files/8199082/Archive.zip)

jamesb93:
Adds the ability to drag a square of an area of the plotter to zoom in.

The shortcut is option + drag for Mac and whatever that translates to on Windows (winkey?)

You can reset the xrange/yrange to whatever was stored via message previously with control + click.

Example can be seen on the final tab of the help file if you want to play with it easily.

### DIFF
--- a/help/fluid.plotter.maxhelp
+++ b/help/fluid.plotter.maxhelp
@@ -37,7 +37,8 @@
 		"tags" : "",
 		"style" : "",
 		"subpatcher_template" : "",
-		"showontab" : 1,
+		"showrootpatcherontab" : 0,
+		"showontab" : 0,
 		"assistshowspatchername" : 0,
 		"boxes" : [ 			{
 				"box" : 				{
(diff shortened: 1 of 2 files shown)


tedmoore (approved):
Nice. I'll put an issue on the SC repo to mirror functionality. Not sure *when* that will happen at the moment. But it'll happen!

jamesb93:
> Nice. I'll put an issue on the SC repo to mirror functionality. Not sure _when_ that will happen at the moment. But it'll happen!

Do you want to await that SC change for the merge?

tedmoore:
> Do you want to await that SC change for the merge?

I think that's a good idea. Should we aim for this to be in beta6?

jamesb93:
> > Do you want to await that SC change for the merge?
> 
> I think that's a good idea. Should we aim for this to be in beta6?

Yep! Possibly, do you think I could tackle the SC version with your supervision and final say on the code style?

tedmoore:
> Yep! Possibly, do you think I could tackle the SC version with your supervision and final say on the code style?

I would like to do it.

jamesb93:
> > Yep! Possibly, do you think I could tackle the SC version with your supervision and final say on the code style?
> 
> I would like to do it.

No problem :) When you make the PR/issue link it up against this one somehow.

tedmoore:
> When you make the PR/issue link it up against this one somehow.

https://github.com/flucoma/flucoma-sc/issues/84

jamesb93:
> > When you make the PR/issue link it up against this one somehow.
> 
> [flucoma/flucoma-sc#84](https://github.com/flucoma/flucoma-sc/issues/84)

🚀

tremblap (approved):

tremblap:
this is great and I'll add an issue on Pd (and let's not wait for it because this object is already breaking with PurrData so I'll have to go radical I think...)

PR #188, opened 2022-05-04
[Docs] Update fluid.list2buf

jamesb93:
Updates fluid.list2buf help file﻿

### DIFF
--- a/help/fluid.list2buf.maxhelp
+++ b/help/fluid.list2buf.maxhelp
@@ -57,7 +57,7 @@
 						}
 ,
 						"classnamespace" : "box",
-						"rect" : [ 0.0, 26.0, 688.0, 455.0 ],
+						"rect" : [ 100.0, 126.0, 688.0, 455.0 ],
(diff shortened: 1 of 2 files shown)


tremblap (changes requested):
looks nice. auto-sizing doesn't shrink the internal buffer - if that is not a bug, we need a comment on the auto-sizing tab saying so.

jamesb93:
> looks nice. auto-sizing doesn't shrink the internal buffer - if that is not a bug, we need a comment on the auto-sizing tab saying so.

I actually don't know and just assumed it was the behaviour of the object. Maybe @weefuzzy can chime in?

tremblap:
I saw the discussion on discord, is that a bug then?

jamesb93:
> An enhancement. IAC, I've said how to enact it, so it could go straight in

If you're happy I shall be your conduit of mighty change.

weefuzzy:
Yay, my son, go forth

PR #204, opened 2022-05-13
[Fix] Various Help File Errors

jamesb93:
Introduced either to badness in loadbang behaviour from Max or my own stupidity.

fluid.pitch~
fluid.ampslice~
fluid.bufampslice~


tremblap:
thanks

PR #214, opened 2022-05-16
[Docs] MelBands + BufMelBands Translation

jamesb93:
As the title suggests.


tremblap:
this is great and thorought and sounds amazing!

tremblap (changes requested):
there is just one error when bufmel is loading:
fluid.bufmelbands~: Input buffer help.mel.1.src invalid (possibly zero-size?)

PR #217, opened 2022-05-17
[Docs] MFCC and BufMFCC Translation

jamesb93:
As it says on the tin, these are the translations of Ted's work.

### DIFF
--- a/help/fluid.bufmfcc~.maxhelp
+++ b/help/fluid.bufmfcc~.maxhelp
@@ -3,7 +3,7 @@
 		"fileversion" : 1,
 		"appversion" : 		{
 			"major" : 8,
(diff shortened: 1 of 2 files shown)


tremblap (changes requested):
great work. some cosmetic/behaviour tweaks:

mfcc: 3rd tab: the easydac starts all tabs and not just that tab

bufmfcc: maybe loading a mono file by default (the piano makes twice as many things)
also the text at the top is cropped.

jamesb93:
Everything addressed thanks!

The cropping will go away with an updated build with -DDOCS=ON due to:

https://github.com/flucoma/flucoma-docs/pull/113

PR #218, opened 2022-05-17
[Docs] BufFlatten Translation

jamesb93:
As it says.


tremblap (changes requested):
the code is great but I am worried of the vocab choice in the new interface. Saying 'without buffer' (and inside explanation too) is misleading. @weefuzzy might have a suggestion on how to say it in a way that makes it clear the buffers are still there, just created for us?

jamesb93:
yeah without buffer is totally misleading.

In other places I have said "automatically managed buffers" or "buffers implicitly created where no output buffer is specified".

jamesb93:
@tremblap I've changed the name of that sub patch to:

image

PR #219, opened 2022-05-18
[Docs] HPSS and BufHPSS translation

jamesb93:
As it says...

- prepare help files with grid/fonts
- cleanup help file layout
- cleanup help files
- update remaining tabs for hpss


tremblap (changes requested):
great work.

bufhpss: tab3 (mode2) the gui are not connected to the object
tab4(mc) the player doesn't play the harmo nor perc

PR #220, opened 2022-05-20
[Docs] BufStats translation

jamesb93:
- prepare bufstats help file
- in progress
- full translation
- add a derivative tab
- cosmetics


tremblap (changes requested):
this is fantastic. there is only a troubleshooting print to remove from the tab 'multichannel' otherwise it is very clear

PR #221, opened 2022-05-20
[Docs] NMFMatch Translation

jamesb93:
- prepare and grid
- work on compressor example
- work in compressor parameters
- finish translation


tremblap (changes requested):
all great except the compressor example that doesn't work anymore - no dry, nor delays...

PR #222, opened 2022-05-20
[Docs] NMFFilter Translation

jamesb93:
As the title says

### DIFF
--- a/help/fluid.nmffilter~.maxhelp
+++ b/help/fluid.nmffilter~.maxhelp
@@ -3,22 +3,22 @@
 		"fileversion" : 1,
 		"appversion" : 		{
 			"major" : 8,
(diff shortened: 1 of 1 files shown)


tremblap (changes requested):
this is good. one error: in tab 1 it seems the ezdac is for all the tabs not just itself. I wonder if we could search the full folder for the attribute name and setting to catch any other such non-local ezdac?

jamesb93:
> this is good. one error: in tab 1 it seems the ezdac is for all the tabs not just itself. I wonder if we could search the full folder for the attribute name and setting to catch any other such non-local ezdac?

Yes indeed we should `ripgrep` for it and squash them all in one hit before release. I'll file an issue so we remember.

PR #226, opened 2022-05-23
[Docs] MLPClassifier Translation

jamesb93:
Translation of Ted's work on the SC help files.

This help file will have a strong link to the videos and a bigger example patch to demonstrate the canonical musical application of audio classification. Having this live outside the help file reduces the bloat and "ARGHHHHH" feeling of opening it up.

These connections don't yet exist because these examples don't but there will be an accompanying issue.

### DIFF
--- a/help/fluid.mlpclassifier~.maxhelp
+++ b/help/fluid.mlpclassifier~.maxhelp
@@ -10,7 +10,7 @@
 		}
 ,
 		"classnamespace" : "box",
-		"rect" : [ 35.0, 88.0, 963.0, 877.0 ],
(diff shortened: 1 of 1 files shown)


jamesb93:
Issue here: https://github.com/flucoma/flucoma-max/issues/227

jamesb93:
On one hand, I think that this help file is messy (again) and not that clear. What do you think @tremblap?

jamesb93:
Should be updated to accomodate changes in:

https://github.com/flucoma/flucoma-docs/pull/109/

tremblap (approved):
good translation a little more dense than the regressor but the 2nd tab makes it a little more fun.

Issue #227, opened 2022-05-23
MLPClassifier and MLPRegressor help files need to link to bigger examples


jamesb93:
Because these objects do so much it bloats the help file to introduce audio classification / parameter regression in them.

To help smooth this over we can link to those example patches within the patch.

jamesb93:
Solved via docs in https://github.com/flucoma/flucoma-docs/commit/a71d467b09f0a9a1fcebf3cac59a3190d576e928

PR #228, opened 2022-05-23
[Docs] MLPRegressor Translation

jamesb93:
Translation of the MLPRegressor SC help file


jamesb93:
This should actually be updated to accomodate changes present in:

https://github.com/flucoma/flucoma-docs/pull/109/

tremblap (approved):
this translation is very dry. I kind of miss the fun synth example although I understand it is too much for a helpfile. this is where the learn page shines with all the cross links and all.

jamesb93:
> this translation is very dry. I kind of miss the fun synth example although I understand it is too much for a helpfile. this is where the learn page shines with all the cross links and all.

Indeed this was the criticism that was presented to me when it used to have the musical example in it. I think that these monolith objects should link out very clearly to more full examples to help bridge the gap in the CCE and with learn. I made an issue here to track it:

https://github.com/flucoma/flucoma-max/issues/227

jamesb93:
> ok that sounds good. the more I think about it the more I think you're right.

Can I merge this then with the follow-up for linking out to bigger tutorials as the plan?

tremblap:
yep

PR #231, opened 2022-05-25
Enhance/generate stubs

weefuzzy:
CMake changes to generate `.cpp` stub files for build from `flucoma-core`'s client list


tremblap:
seems fine - compiles now on macos - untested beyond make install

PR #233, opened 2022-05-27
[Docs] Standardise SC Translation

jamesb93:
As the title says.


tremblap:
it works. It is a bit sad that we lost the graphic representation (which I reckon we lost in SC too @tedmoore ) but it is probably more readable. The reference online is still empty, is that normal?

tedmoore:
> It is a bit sad that we lost the graphic representation

I'm not sure what you're referring to. Perhaps it wasn't in the SC example code?

tremblap:
it was: there was a bipolar bar graph for the raw and scaled values. I remember coding it influenced by @weefuzzy 's Max version (translation was in that direction then )

tedmoore:
> it was: there was a bipolar bar graph for the raw and scaled values. I remember coding it influenced by @weefuzzy 's Max version (translation was in that direction then )

Ah, I remember now. Does "it is a bit sad" mean that you want for it to be put back in?

tremblap:
> Ah, I remember now. Does "it is a bit sad" mean that you want for it to be put back in? I'm already on it

it was more to raise a question - was it removed because it was considered confusing (so leave it out) or for another reason?

jamesb93:
> > Ah, I remember now. Does "it is a bit sad" mean that you want for it to be put back in? I'm already on it
> 
> it was more to raise a question - was it removed because it was considered confusing (so leave it out) or for another reason?

It was confusing to me which was my rationale for not questioning it in the review of the SuperCollider pull request.

tedmoore:
> > > Ah, I remember now. Does "it is a bit sad" mean that you want for it to be put back in? I'm already on it
> > 
> > 
> > it was more to raise a question - was it removed because it was considered confusing (so leave it out) or for another reason?
> 
> It was confusing to me which was my rationale for not questioning it in the review of the SuperCollider pull request.

I don't think it's too confusing. I found it a little dry and thought that it didn't really add much. But I don't think it's hurting anything. Generally my approach has been the more angles we can show on things the better.

PR #234, opened 2022-05-27
[Docs] RobustScaler SC Translation

jamesb93:
As the title says


tremblap (changes requested):
this is good. again the first tab has lost its graphic representation but that is probably deliberate.

the scaler comparison in tab 2 is great and could be duplicated in all 3 scalers maybe (programmatically) and it would be good if we could zoom all 3 manually (the -10 10 range is a bit difficult to grasp) and I think they should all zoom together and maybe starting with -3 3

PR #241, opened 2022-05-31
[Docs] AmpGate SC Translation

jamesb93:
- grid help file
- translate ampgate

### DIFF
--- a/help/fluid.ampgate~.maxhelp
+++ b/help/fluid.ampgate~.maxhelp
@@ -58,7 +58,7 @@
 						}
 ,
 						"classnamespace" : "box",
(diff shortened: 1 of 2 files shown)


tremblap (changes requested):
there is a conceptual error in the player - I love the interleave but it means 0 plays the first slice, 1 plays the first silence, which is not super clear. either we keep it like this and explain a bit more verbose, or we remove the interleaving and play from onset to offset only?

tab2: the view in SC is better (it selects the onset to the offset) I wonder if that is a feature of waveform or if @tedmoore has made a custom transform?

tab3: the loadbang doesn't seem to fire so we don't see any waveform

tab4 (removing silence) is trippy to listen to in glorious stereo :)

PR #245, opened 2022-06-06
[Docs] Stats SC Translation

jamesb93:
- grid it out
- update first tab
- mimic Ted's real use case tab


tremblap (approved):
man you guys are on fire. I could look at that first tab for ages. nice one!

PR #247, opened 2022-06-06
[Docs] PCA SC Translation

jamesb93:


tremblap (approved):
great - just a bit sad that the class colours don't match in the whitening examples that would make it even clearer

PR #248, opened 2022-06-06
[Docs] STFT SC Translation

jamesb93:
- prepare help file
- rework first tab
- in progress
- cleanup first tab
- musical stft example
- mild cleanup
- finish musical example


tremblap (approved):
great example - the buffer doesn't autoload but it is a non-issue. I'll study that patch.

PR #251, opened 2022-06-08
[Docs] BufNNDSVD SC Translation 

jamesb93:


weefuzzy:
> one error on load: fluid.bufnndsvd~: "numchans" is not a valid attribute argument - which is missing it seems. @weefuzzy were we aware of that?

We were, in as far as you can see that the client has never had `startFrame` etc.

PR #259, opened 2022-06-10
[Docs] Transients Family Translation 

jamesb93:
- translate transientslice
- translate buftransientslice
- fluid.transients~ translation
- cosmetic
- translate buftransients

### DIFF
--- a/help/fluid.buftransientslice~.maxhelp
+++ b/help/fluid.buftransientslice~.maxhelp
@@ -3,22 +3,22 @@
 		"fileversion" : 1,
 		"appversion" : 		{
 			"major" : 8,
-			"minor" : 1,
+			"minor" : 3,
 			"revision" : 0,
 			"architecture" : "x64",
 			"modernui" : 1
 		}
 ,
 		"classnamespace" : "box",
-		"rect" : [ 100.0, 100.0, 673.0, 527.0 ],
+		"rect" : [ 100.0, 100.0, 846.0, 626.0 ],
 		"bglocked" : 0,
 		"openinpresentation" : 0,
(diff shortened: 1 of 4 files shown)


tremblap (commented):
these are ok. buftransient could be improved in 2 ways

- as discussed with @tedmoore for the SC one, we should look at 2 seconds instead of a long file since looking at it like this shows very little

- I prefer the stereo example to be on dual mono as it illustrates more the independance of the channels (the old guit+piano files was more explicit despite 'costing' a bit bufcompose.

jamesb93:
> * as discussed with @tedmoore for the SC one, we should look at 2 seconds instead of a long file since looking at it like this shows very little

image

I think this pretty clear compared?

> * I prefer the stereo example to be on dual mono as it illustrates more the independance of the channels (the old guit+piano files was more explicit despite 'costing' a bit bufcompose.

I agree!

tedmoore:
I remember a conversation with @tremblap that was rather than having stereo examples in all the example codes we should make a stereo processing overview somewhere...I've been deleting all the stereo processing examples from the code (:eyes:)

jamesb93:
I agree it would be much easier to maintain and deal with to be honest.

I'm on board for the idea of shared tutorials/references/guides over patch specific ones. It would be important to caveat the difference between everything else and the `BufNMF` behaviour though (and possibly related objects but I don't remember off the top of my head).

tedmoore:
> It would be important to caveat the difference between everything else and the `BufNMF` behaviour though

Can you be more specific about what the difference is in this case? 

In general I have been checking that the RST has information about "what happens when a multichannel file is used" which they do.

PR #264, opened 2022-06-13
skmeans translation from SC (missing one tab also missing in kmeans)

tremblap:


tremblap:
@g-roma if you look at the 2nd tab, it seems that the algo is converging very quickly towards 2 clusters only. the visualisation is quite good in Max to help me understand I don't understand it yet... or there is a bug... can you shed some light on that behaviour?

tremblap:
actually this is a bug in all CCEs I'm exploring and will post in the general PR on core.

Issue #224, opened 2022-05-21
auto-named `dataset~`s (et al) are not referrable 


rconstanzo:
After a long while of head scratching trying to figure out why my patch isn't working, it turns out that it's not possible to `refer` a name onto a `fluid.dataset~` that exists inside an abstraction/subpatch.

I compared the behavior with `jit.matrix`, and it seems like `jit.matrix` has no problem "seeing" the name from inside an abstraction.

This is especially poignant as the solution to concatenating `fluid.dataset~`s [will likely involve creating abstractions](https://discourse.flucoma.org/t/concatenation-in-the-select-based-future/1439) to manage the shuffling around of `fluid.dataset~`s to get them all joined up.

![Screenshot 2022-05-21 at 9 58 09 pm](https://user-images.githubusercontent.com/2747015/169668818-68eeae60-e23e-4f62-b268-e913fb488e7e.png)

[example_code.zip](https://github.com/flucoma/flucoma-max/files/8748338/example_code.zip)

rconstanzo:
It's slightly more peculiar. If I get the name from `getname`-ing the abstraction part, and then manually name the top-level `fluid.dataset~` (e.g. `fluid.dataset~ u818009803`) then it works. It seems that the `refer` message, specifically, can't see inside the abstraction.

jamesb93:
I'll have to verify for myself, but what you are saying is a `fluid.dataset~` doesn't seem to know about other `fluid.datasets~` outside of their patcher?

rconstanzo:
Well, it does if you manually name the object (e.g. `fluid.dataset~ datasetName`),  but if you `refer datasetName` instead, it doesn't like it.

There's a code tidbit there, it's just zipped because of the need for having abstractions in the example.

This one is quite annoying to code around as there's no good way around it (as far as I have figured out), so I end up `getname`-ing the dataset, and then manually re-instantiating the `fluid.dataset~` I want to use... every time I'm editing/using the patch.

weefuzzy:
This isn't to do with whether there's an abstraction or not; it's simply a logical error in how the auto-naming case gets handled which means that auto-named shared objects aren't making it into the lookup table properly.

jamesb93:
@weefuzzy has fixed the issue in 224 :)

PR #297, opened 2022-06-21
Issue 224

weefuzzy:
fixes issue #224 

Ensures that nameable data objects (dataset et al) that are auto-named get properly placed in lookup tables


jamesb93:
Appears to be fixed using Rod's test code.

PR #300, opened 2022-06-21
[docs] guides

jamesb93:
This adds a guide for automatic `buffer~` and `dataset~` management. This includes a reference entry under "Guides" as well as a patch. Alongside this every buffer~ processor and every dataset~ processor now have a shared tab, that takes them to this patch.

I would appreciate your okay on this. I am fairly confident that the content is solid as @weefuzzy already took a glance.

@weefuzzy Would you be able to test the install changes I made to `cmake`. They may not be correct but seemed fine my end.


weefuzzy:
install seems to work fine, yes

weefuzzy (approved):

tremblap (approved):
this looks great and compiles and links work. I was surprised of the location of the files (docs/vignettes/flucoma) but I reckon that is the right place since that works ;)

jamesb93:
> this looks great and compiles and links work. I was surprised of the location of the files (docs/vignettes/flucoma) but I reckon that is the right place since that works ;)

Yeah its a little unusual but I think, from what I can tell, that folder there dictates the "module" that the vignette belongs to. HIRT does this and so does a lot of the internal packaging system so that you can have vignettes with the same name across different parts of the docs... hard to tell as none of this is documented though.

PR #337, opened 2022-12-06
Peaks

tremblap:
This is the helpfile in progress for (buf)sinefeature

Missing from the wrapper, I reckon, are ways to deal with the 2 outputs (freq and mag)

1. for the RT version, there is only freq coming out, and dumpout is duplicated

EDIT: the buff version works as expected

### DIFF
--- a/extras/Fluid Corpus Manipulation Toolkit.maxpat
+++ b/extras/Fluid Corpus Manipulation Toolkit.maxpat
@@ -3,8 +3,8 @@
 		"fileversion" : 1,
 		"appversion" : 		{
 			"major" : 8,
-			"minor" : 5,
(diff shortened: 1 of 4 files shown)


tremblap:
(this needs the peak PR in core and docs)

tremblap:
and the BUF output is very strange compare to the SC one - loads of noise despite a simple input of 3 sinewaves - even when fft is put to 2048. compare with SC example (same freqs and amps in both helpfiles)

weefuzzy:
Buffer output looks interleaved to me, which would tally with the NRT wrapper apparently taking no account of multiple control outlets that I can see. What did we do in SC to make this apparently work

Issue #335, opened 2022-10-05
Typo in universal `blocking` tab


rconstanzo:
The digest for the blocking tab across all the objects says:
`Choose with thread to use for processing`

I imagine that is meant to be a "which".

jamesb93:
Fixed in https://github.com/flucoma/flucoma-max/pull/339.

PR #339, opened 2022-12-09
fixes blocking tab typo

jamesb93:
Fixes issue linked in https://github.com/flucoma/flucoma-max/issues/335.


tremblap:
thanks for the fix @jamesb93 !

PR #357, opened 2023-02-21
replace `set-output` with environment file

weefuzzy:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


jamesb93 (approved):
Looks good. Let's see if it behaves...

Issue #371, opened 2023-08-02
`fluid.list2buf` fails build with negative subscript error


lewardo:
When compiling on Windows 10 and 11 with Visual Studio Tools 2022 and Max SDK version 8.2.0,  fluid.list2buf fails to compile, only for debug targets

```
2>fluid.list2buf.cpp
2>C:\Users\lewar\Documents\flucoma\flucoma-max\source\include\clients/nrt/FluidListToBuf.hpp(88,3): error C2118: negative subscript
2>C:\Users\lewar\Documents\flucoma\flucoma-max\source\include\clients/nrt/FluidListToBuf.hpp(125,36): warning C4244: 'argument': conversion from 'long' to 'short', possible loss of data
2>C:\Users\lewar\Documents\flucoma\flucoma-max\source\include\clients/nrt/FluidListToBuf.hpp(142,29): warning C4244: 'argument': conversion from 'long' to 'short', possible loss of data
2>C:\Users\lewar\Documents\flucoma\flucoma-max\source\include\clients/nrt/FluidListToBuf.hpp(256,69): warning C4244: 'return': conversion from 't_atom_float' to 'float', possible loss of data
2>Done building project "fluid.list2buf.vcxproj" -- FAILED.
```

traced into `fluid.list2buf.cpp:88`
```
 CLASS_ATTR_SYM(FluidListToBufClass, "destination", 0, FluidListToBuf, outName);
```

which expands to
```
#define CLASS_ATTR_SYM(c,attrname,flags,structname,structmember) \
		{		\
			C74_STATIC_ASSERT(structmembersize(structname,structmember)==sizeof(t_symbol*), "structmember must be t_symbol* type"); \
		class_addattr((c),attr_offset_new(attrname,USESYM(symbol),(flags),(method)0L,(method)0L,calcoffset(structname,structmember))); \
		}
```

so assuming the error is in `class_addattr((c),attr_offset_new(attrname,USESYM(symbol),(flags),(method)0L,(method)0L,calcoffset(structname,structmember)))` given error message emitted

looking on the Max issues this bug has already been encountered in [https://github.com/Cycling74/max-sdk/issues/54](https://github.com/Cycling74/max-sdk/issues/54)

The issue was closed and seems to be an issue with the max-sdk update from 8.0.3 to 8.2.0, they updated the macro for their type, but my knowledge of the SDK isn't refined enough to know the equivalent fix since their error was a float/double CLASS_ATTR rather than a t_symbol

[...]

PR #372, opened 2023-08-02
Fixes issue #371 - wrong casting type for outName

fearn-e:
As seen in https://github.com/Cycling74/max-sdk/issues/54 where this same bug was encountered, it seems like this is an error caused by using the wrong static assert for the outName structure component of the FluidListToBuf struct.

In the original code, a CLASS_ATTR_SYM is used:
`CLASS_ATTR_SYM(FluidListToBufClass, "destination", 0, FluidListToBuf, outName);`

However, this static assert creates an attribute of type t_symbol, and outName is of type t_atom:
```
struct FluidListToBuf
{
  t_object      obj;
  t_object*     defaultOut;
  t_buffer_ref* outputRef;
  t_symbol*     defaultOutName{nullptr};
  t_atom        outName;
  index         axis{0};
  index         canResize;
  index         startChannel{0};
  index         startFrame{0};
  Buffer        output;
  void*         outlet;
};
```

Changing the original line to use a CLASS_ATTR_ATOM seems to solve this issue.

### DIFF
--- a/source/include/clients/nrt/FluidListToBuf.hpp
+++ b/source/include/clients/nrt/FluidListToBuf.hpp
@@ -32,7 +32,7 @@ struct FluidListToBuf
   t_object*     defaultOut;
   t_buffer_ref* outputRef;
   t_symbol*     defaultOutName{nullptr};
-  t_symbol*     outName;
+  t_atom        outName;
   index         axis{0};
   index         canResize;
   index         startChannel{0};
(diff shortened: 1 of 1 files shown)


junofern:
RE: https://github.com/junofern/flucoma-max/commit/d410b7eeda5861c6cb09b0f3605f536af2d51fae#commitcomment-123601134 @weefuzzy
Just changing the macro does work perfectly fine, so I've reverted back to just that change for this PR. This should now be a nice clean and simple fix. I'll open an additional PR regarding changing the type of outName to a symbol.

PR #375, opened 2023-08-03
Added missing patch chord to fluid.audiotransport~ help file second tab

MattS6464:


AlexHarker:
Nice to see this PR - If you look at the diff (changed lines) you'll see a large block that is about a "dependency cache" - that's file paths that Max stores locally to make reloading quicker. Some of those paths will be local to your computer.

It's best to stop max from doing this when committing to repos. You can do this by going into Max Preferences and turning "Save Dependency Paths" off. Once you've done that, resave the patch you've altered here and make a new commit "remove dependency paths" or similar.

PR #379, opened 2023-08-07
EOL `.gitattributes` normalisation

lewardo:
EOL standardisation with `.gitattributes` to fix issues with versions of Git before 2.10 allowed automatic EOL conversion.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,17 @@
+* text=auto
+
+*.cmake text
+*.cpp text
+*.gitattributes text
+*.gitignore text
+*.hpp text
+*.in text
+*.js text
+*.maxhelp text
+*.maxpat text
+*.md text
+*.txt text
+*.xml text
+*.yml text
+
+*.png binary


weefuzzy:
Thanks for the PR. Not sure if text type really is best for `maxpat` and `maxhelp` files because line-by-line diffs don't make sense for these formats – probably better to mark them binary.

jamesb93:
I would err on the side of treating them as text, as is the default position of `git` anyway. If in the future we have more sensible ways of diffing the files, not having to update this file is a win. In the end it doesn't really matter either way, because the given situation is that diffing does nothing appreciable for these two file types.

AlexHarker:
If the maxpat is correctly sorted (which I think was fixed last year sometime in Max) then the diffs make sense, but are often dense.

If max coders are disciplined they are perfectly readable for small changes. For larger changes it gets hard, and margining is still a terrible idea, but I can usually understand maxpat changes from the commit diffs only if they are small edits.

weefuzzy:
> If the maxpat is correctly sorted (which I think was fixed last year sometime in Max) 

Oooh, didn't realise that had happened. Praise be. 

> If max coders are disciplined 

I'll leave that there for our future shared belly laughs.

tremblap:
ok so should I merge in all repos with these settings? there is a PR on all of them.

PR #389, opened 2023-08-14
Fluid.jit.plotter

balintlaczko:
Added the `fluid.jit.plotter` abstraction based on `jit.pworld`.

# Motivation

While `fluid.plotter` is great for most use cases, its performance suffers considerably with large datasets (north of 100K points). Out of curiosity I checked how a `jit.pworld` with `jit.gl.mesh`-es compared and it turned out they were much faster – so I set out to make `fluid.jit.plotter`.

# Interface – Parity

I tried to copy the interface & behavior of `fluid.plotter` as much as I could, however there are a few key differences.

## `refer`

To "load" a `fluid.dataset~` (left inlet) or a `fluid.labelset~` (right inlet) you need to use the `refer` message instead of the `dictionary` (resulting from a `dump` on the dataset/labelset). The motivation for this was that loading very large datasets with the dictionary `dump` method can be slow, and using `refer` saves one round of copying. Upon testing I could verify that datasets with 1M points load much faster via `refer` than via dictionary-dumping.
However (sadly) the right inlet still can listen to `dictionary` as it still uses dictionary-dumping under the hood. I changed the interface to `refer` anyway for consistency (though, again, `dictionary` can still be used) in the hope that I'll find a way in the future to determine the number of unique labels in a labelset without needing to iterate.

UPDATE: now the left inlet also supports `dictionary` as suggested by @jamesb93, so now the abstraction can function as a drop-in replacement for `fluid.plotter` (for the most part, read on...).

## `pointsize` and `highlightedpointsize`

Instead of the `pointsizescale <float>` message `fluid.jit.plotter` uses `pointsize <int>` for setting the size of non-highlighted points and `highlightedpointsize <int>` for setting the size of highlighted points. The type change to int is because under the hood it is setting the `point_size` of the two `jit.gl.mesh`-es (one for the highlighted points, and one for all points). I thought handing over the actual controls to the meshes is more "truthful" than implementing a higher-level `pointsizescale` on top of them. This also allows to decouple the two sizes from each other.

## No individual point adding/sizing/coloring

Since I really wanted to optimize `fluid.jit.plotter` for high query speed I had to sacrifice `fluid.plotter`'s `setpoint`, `pointcolor` and `pointsize` low-level drawing commands. The individual point sizing is not possible (to my knowledge) with `jit.gl.mesh` as of writing this (or only through _very_ involved hacks with depth and camera lens projection), and while adding and coloring points individually might technically be possible, it would add another source of truth for point-colors (than the label-based color matrix) and that would mean an additional check for each highlighted point when querying that I wanted to avoid in the name of speed. Adding points with `setpoint` is the least prohibitive, but without the individual sizing and coloring I did not see much use in it (and again, wanted to avoid multiple sources of truth).
It would technically be possible to have the individual size/color (even shape) if I'd use instancing with `jit.gl.multiple` instead of a `jit.gl.mesh`, however updating the attribute matrices of a `jit.gl.multiple` (for example when changing the size of _one_ point) get _very_ prohibitive with large datasets, and the whole idea behind `fluid.jit.plotter` was to have fast querying with large datasets (and lots of highlighted points).
Right now points must always come from a reference to a dataset, and colors must always come from a reference to a labelset.

## The rest is the same

Everything else aims to behave the same way as `fluid.plotter`, that is:
- `highlight`
- `xrange`, `yrange` and `range`
- `shape` and `bgcolor`
- `colorscheme`
- `clear` and `reset`
- click’n’drag zooming and reporting `zoomxrange` and `zoomyrange` via the 2nd outlet
- being resizable and forcing (somewhat jankily) a square aspect ratio

# Conclusion

`fluid.jit.plotter` is a variant of `fluid.plotter` that is aimed towards speedy querying with large datasets. For this it is based on `jit.pworld` instead of `jsui` and sacrifices some flexibility in terms of its appearance. It is really intended for those who need fast querying with datasets with hundreds of thousands, or millions of points. For smaller datasets `fluid.plotter` is still preferable due to its additional flexibility.


jamesb93:
Thanks for this hefty PR @balintlaczko. This is really an exemplary contribution – docs, clean code, very clear motivation. Nice!

My first impression is that this is a great addition to the toolkit. Ive seen a lot of `fluid.plotter` in the wild (Taylor Brook's videos come to mind) and offering a "batteries included big data fast version" is cool. I would say that we should emphasise this difference whenever the next release comes about. In general though the new abstraction feels a lot snappier than the original `jsui` plotter. This is a win in any case and offers people on slower machines the ability to do more with the toolkit.

For me to say 🚀 APPROVE 🚀, I would want to ensure that `fluid.plotter` and `fluid.jit.plotter` are as interchangeable as possible within reason.

## refer

In principle I like `refer` even though it diverges from `dictionary`. I noticed from trying out your code that the new `fluid.jit.plotter` doesn't accept `dictionary` as a message selector. Can we alias `refer` to `dictionary` so that existing patches work without any changes? 

## clear

I also noticed `clear` is no longer a valid message. Let's alias `clear` functionally to calling `refer` without any arguments

## Big Data Documentation

image

The typical style guide is that calls to action (numbered steps) should be inside a comment box bubble. See the first tab and copy that style.

Other than these points I have no reservations. It looks really good! I say we make this a part of the next release, whenever that is and the nightlies ASAP.

javascript/jit_plotter_labels2colors.js
```diff
@@ -0,0 +1,72 @@
+autowatch = 1;
```
jamesb93, line 1: I would rename this file `fluid.jit.plotter.labels2colors.js` to keep it namespaced and consistent.

javascript/jit_plotter_resize_pworld.js
```diff
@@ -0,0 +1,77 @@
+// create a Task event loop detecting bpatcher resizing
```
jamesb93, line 1: I would rename this file `fluid.jit.plotter.labels2colors.js` to keep it namespaced and consistent.

jamesb93 (changes requested):

javascript/jit_plotter_resize_pworld.js
balintlaczko: you mean `fluid.jit.plotter.resize.pworld.js`

local_docs/fluid.jit.plotter.maxref.xml
```diff
@@ -0,0 +1,220 @@
...
+Part of the Fluid Corpus Manipulation Project (http://www.flucoma.org/)
+Copyright 2017-2019 University of Huddersfield.
+Licensed under the BSD-3 License.
+See license.md file in the project root for full license information.
+This project has received funding from the European Research Council (ERC)
+under the European Union’s Horizon 2020 research and innovation programme
+(grant agreement No 725899).
+-->
+<c74object name='fluid.jit.plotter' category='FluidCorpusManuipulation'>
+    <digest>Plot data in a two-dimensional graph.</digest>
```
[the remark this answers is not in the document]
balintlaczko: changed to "Plot very large 2D datasets fast."

tremblap:
It'll be part of my checks later this week

balintlaczko:
> Thanks for those changes @balintlaczko , we can also be sure this all works on Windows before we merge?

Tested on Windows 11, Max 8.5.5, works great and snappy!

jamesb93 (approved):

tremblap:
I am not certain where it will live yet, so do not merge until I've approved.

tremblap (changes requested):
ok first, congrats on all of this. It is fantastic. the 1M points example is mindblowing. 

A few points: 

1. I agree with @jamesb93 that dictionary should replace refer (do we actually need refer at all) - actually any difference in underlying behaviour is trivial so we should copy the same interface and leave a note in the doc

2. I don't know if the object should have its own helpfile or just a single tab in the fluid.plotter - I'm in two minds - one thing for sure, if it is really stand alone, we really need to add a tab to point at it in fluid.plotter!

Congrats again!

balintlaczko:
Thanks, @tremblap for the kind words! :) I like the idea of fusing the help with `fluid.plotter`, but then we might have to add 2-3 new tabs to it, and the differences in the interface might become more confusing though (like the meaning of the `pointsize` message..).

As for the refer vs. dictionary issue, as of now it supports both, so I can change the documentation to only discuss `refer` as a useful alternative for large datasets.

> (do we actually need refer at all)

Here is a patch that I hope will convince you why is it useful to have `refer`:
image
[refer_vs_dictionary.maxpat.zip](https://github.com/flucoma/flucoma-max/files/12395564/refer_vs_dictionary.maxpat.zip)

In the example I take the generated 1M dummy dataset and read it into two `fluid.jit.plotter`-s. The left one uses `dictionary` dumping, the right one uses `refer`. Roughly speaking it is 20 seconds vs half. (Another cold-start result was 20176 ms vs 561 ms., third one was 20044 vs 585.) That is a big time difference to wait every time you load a dataset. I haven't explicitly test memory there, but considering that in the `dict` dump method the whole of the dataset continues to exist both as a fluid dataset _and_ as a dictionary it seems a bit prohibitive for large datasets. The whole motivation for making `fluid.jit.plotter` has been to make a `fluid.plotter` that is optimized for speed and performance with large datasets, that's also why I decided to move to using `refer`.

What if we rebrand `refer` as a good alternative interface for working with large datasets (perhaps with the attached patch as a tab that shows the diff) and otherwise keep `dictionary` as the expected interface? That would mean me redoing the help and the reference a bit, otherwise everything could stay the same? What do you think?

tremblap:
> > (do we actually need refer at all)

Sorry I wasn't clear - I meant we should not have the refer message but always do refer under the hood. we don't need a copy, do we?

balintlaczko:
Aha, I see what you mean now! I see two problems with that:
1. The dictionary is usually "anonymous", and you can't infer the source dataset it came from form its name – as far as I know.
2. The bigger problem is that we even render dictionaries of large fluid datasets. This is in fact what takes most of the time. 
image
 
[refer_vs_dictionary_2.maxpat.zip](https://github.com/flucoma/flucoma-max/files/12406951/refer_vs_dictionary_2.maxpat.zip)

Internally, an incoming dictionary goes to a "private" `fluid.dataset~` using `load` and then we `refer` to that private dataset same as we would an external one.
image

> we don't need a copy, do we?

Exactly my point! But to keep using the `dump` method of `fluid.dataset~` while avoiding copying would be if the object would already manage its data internally in a Max `dictionary` so that upon "dump" it would just emit the name of the dictionary. But I guess this is dangerous for many other reasons...

balintlaczko:
OK, added the new panning mousing mode, but hold on a bit, please, until I test it more thoroughly and document it in the help + reference (there is a new `panfactor` message for unnatural people).

https://github.com/flucoma/flucoma-max/assets/45127463/60def580-2b35-47b2-889d-ed3163d3a67c

tremblap:
ok, this is genius! thanks so much. 3 small changes requested:

1. can you explain why refer and not dump beside the message in the first tab - it is meant for advanced users but knowing cost of copy is important to seed there I think.
 
2. can you please add the [p ?] in the root of the helpfiles for both plotter and jit.plotter? we seem to have forgotten it in the former, and that autopopulates with the references and see also.

3. can you add a tap in the help of plotter entitled large dataset (or somethng) and give an example of 1000 random points for both with timers (like you did to convince me) and add the red text (not in red) of the disclosure of what is different so people can go there with the right expectations (speed at the cost of feature flex is convincing and is well explained in the ref.

congrats again!

balintlaczko:
> can you explain why refer and not dump beside the message in the first tab - it is meant for advanced users but knowing cost of copy is important to seed there I think.

image

> can you please add the [p ?] in the root of the helpfiles for both plotter and jit.plotter? we seem to have forgotten it in the former, and that autopopulates with the references and see also.

Done!

> can you add a tap in the help of plotter entitled large dataset (or somethng) and give an example of 1000 random points for both with timers (like you did to convince me) and add the red text (not in red) of the disclosure of what is different so people can go there with the right expectations (speed at the cost of feature flex is convincing and is well explained in the ref.

image

This is my proposal, but happy to change anything!

balintlaczko:
Okay, tested on Windows again, everything seems fine.

tremblap:
thanks so much for doing this. One last detail: in the plotter help patch, are they both jit.plotters as written and the top one uses a copy instead of refer? If so we need 3: fluidplotter, fluid jitplotter using dump and fluid jitplotter using refer.

if it is an error, we need to change the text of (3) to 
Pass and copy the dictionnary via dump 
and of the (4) to
Pass the name of the dataset to refer to without copying

or somethig more english but stressing super explicitely the difference for the quick reader in the main bubbles

do you agree?

balintlaczko:
> (oh and it says there is a conflict with fluid.plotter help - don't forget to resolve the conflicts)

I think I did that now, it was just 2 "rect" values and the "saturation" of the [swatch]. The latter probably changes every time you save the help.

balintlaczko:
> are they both jit.plotters as written and the top one uses a copy instead of refer?

Yes, they were both `fluid.jit.plotter`s. I made the top one into a `fluid.plotter` and changed the comments to your suggestions. I changed it slightly, so that the 2nd bubble ends with "via dump". So it's like:
2) Render a dictionary from the dataset via 'dump'
3) Pass and copy the dictionary to fluid.plotter
4) Pass the name of the dataset without copying to fluid.jit.plotter

I think the mention of copying is fine (I guess it _does_ copy the dictionary contents upon `load`ing it to the "private" dataset), but I also wanted to emphasize in this tab that the practice of rendering the dictionary using `dump` _is_ the main bottleneck with large datasets. At this scale the timings don't show that explicitly: with 10K points it is around 2/3 of the time to render the dictionary and 1/3 to load from it, but with [1M points](https://github.com/flucoma/flucoma-max/pull/389/#issuecomment-1687776719) you see that the proportions change drastically. 

But maybe this discussion is too much to cram into a help file, so it's enough to indicate that with `refer` we avoid copying. (And as a side effect we also avoid the dictionary-rendering which is actually the bigger speed bump.)

image

balintlaczko:
PS: keep suggesting changes if there is anything you don't like, I know making help patches is a delicate business...

tremblap:
ok I'll play with them as is - I think this is good for a merge - if 1M points is better maybe 100k is a good compromise?

balintlaczko:
> if 1M points is better maybe 100k is a good compromise?

Yes, I think 100K is a good compromise, it's already a large dataset size, and the ratio there is already roughly 7 : 3 for dictionary rendering : loading on my machine. But the curve seems exponential, because with 1M points the ratio is roughly 10 : 1. But then again, rendering 1M points might cause some serious hangs on older machines, so let's stick with 100K. There is already a 1M example tab in the `fluid.jit.plotter` help, I guess that's enough.

balintlaczko:
Updated:
image

tremblap (approved):

PR #390, opened 2023-08-15
Added NMF example patches to Toolkit patch

MattS6464:
The JIT-NMF example tab (changed to just NMF) now opens a patch with links to four nmf example patches that have all been commented and given descriptions.


tremblap:
Great work!

I think the tone is a bit brutal in the sentence
"If you lack a basic understanding of the nmf objects, check out their help files before diving deeper!"

Other grammar and wording questions:
"in close to realtime" -> close-to-real-time or close-to-realtime ?

in 'resynthesis' there is an extra period after [buf.nmf~]

otherwise I love this very much - and the explainers in the various patches are good! Maybe @weefuzzy or @AlexHarker might want to read the explanations to make sure he likes the inviting tone as much as I do?

tremblap (changes requested):
As written above, a general check of English language (tone and invitation and accuracy of explanation) would need a native English @AlexHarker or @weefuzzy or ...)

PR #397, opened 2023-08-22
Rework Max wrapper to allow multiple inputs and different IO sizes fo…

AlexHarker:
…r control objects

This work supersedes other work towards making the voice allocator object work

**What it does**
- allows the wrapper to use multiple inputs for different things (this work was in a voice allocator branch, but it is presented here in an edited form
- allows for two different ways of setting input size, one of which is dependent on input size and one of which is not - the difference is determinable at compile time / without a client (this relies on a related PR in core (flucoma/flucoma-core#248)

**Changes**
- I removed the deferral from high priority thread for resizing based on input. There are other places in max (the app) where allocation happens in high priority and it seemed overly fussy/complicated - the restriction with buffer~s is c74 imposed,
- even in the wrapper already I can see places where high priority allocation can occur, so it seemed weird to insist on not doing it in this one place (given all the added indirection and logic required to do so).
 
**Notes:**
- functionality for both the bufstats and voice allocator objects should be correct, and @tremblap and I have determined that these should be the only affected objects.
- similar changes are required in other repos (at least pd, although SC and CLI should be considered also).

Sorry to spam @weefuzzy but I suspect that the removal of the deferral stuff is most contentious here. Also apologies that the PR isn't easier to follow, but there was no obvious way to break this work up.

### DIFF
--- a/source/include/FluidMaxWrapper.hpp
+++ b/source/include/FluidMaxWrapper.hpp
@@ -588,19 +588,29 @@ class FluidMaxWrapper
     void operator()(FluidMaxWrapper* x, long ac, t_atom* av)
     {
       FluidContext c;
-            
+
+      index whichIn = proxy_getinlet((t_object *)x);
+
       atom_getdouble_array(std::min<index>(x->mListSize, ac), av,
                            std::min<index>(x->mListSize, ac),
-                           x->mInputListData[0].data());
-      x->mClient.process(x->mInputListViews, x->mOutputListViews, c);
+                           x->mInputListData[whichIn].data());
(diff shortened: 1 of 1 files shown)


AlexHarker:
It's not ideal to allocate on a high priority thread, but I think in the max source you would likely find loads of places where it happens (e..g. attribute getters) all the time, so it seems churlish to worry about it here.

Likewise it happens in  outputMessage() [several if not all variants] and invokeMessageImpl(), because the memory for std::vector will be on the heap, even though the vectors themselves are local....

PR #406, opened 2023-09-06
fluid.voiceallocator initial docs

fearn-e:
@tremblap 
Overview patch update and added help patch for fluid.voiceallocator with a basic tab and an extra musical example tab.


MattS6464:
Added a bufvoiceallocator help patch with an audio example. Also made some slight layout tweaks to the rt voiceallocator help patch.

PR #410, opened 2023-09-12
New example patches

MattS6464:
Added two new example patches to the examples folder.

One uses a series of buffer processing objects like nmf and hpss to decompose a drum loop to be used in a granular synth.
The other uses fluid.pitch~ at the output of a delay line and uses the pitch/confidence values as modulation parameters for the delay, creating a sporadic flanging effect.

These patches have not been added to the Toolkit patch


tremblap:
these need adding to the overview (the last sentence is cryptic :) )

PR #417, opened 2024-04-10
Fluid.jit.plotter update - full parity with fluid.plotter

balintlaczko:
This is a major update for the abstraction, as it finally achieved full parity with the interface and behavior of `fluid.plotter`, while keeping the speed benefits.

image

# Added
- `pointcolor` message for setting individual point colors (parity with the same message of `fluid.plotter`). This can be especially useful when every point needs a unique color associated with it:
image

- `setpoint` message for setting individual points, just like in `fluid.plotter`. There is a caveat though, since `fluid.jit.plotter` also has the `refer` interface to load large datasets super fast. To avoid accidentally modifying outside datasets, if an earlier reference was specified (via the `refer` message) the plot will switch to its internal dataset, making the previous set of points disappear. If you want to load a dataset and then edit its points via the `setpoint` message, then load the dataset first as a dictionary.

- `pointsize` will now allow to change the size of individual points (minor compatibility break to the previous `fluid.jit.plotter` version). Also, there is no `highlightedpointsize` message anymore.

- `pointsizescale`, that will scale all points (regular and highlighted), similar to `pointsizescale` in `fluid.plotter`.

- `highlightscale`, that allows you to change the size ratio between the highlighed points and regular ones; defaults to 2.4.

# Fixed
- zooming in now respects implicit mirroring via `range` (bug demonstrated [here](https://discourse.flucoma.org/t/pd-plotter-zoom-road-test-community-interface-change-question/2161/16?u=balintlaczko))

- decoupled object position between Edit mode and Presentation mode (bug reported [here](https://discourse.flucoma.org/t/fluid-jit-plotter-for-a-better-performance-than-fluid-plotter/1884/51?u=balintlaczko))

- now using `jit.gl.render` instead of `jit.pworld` to render frames on-demand, instead of a constant 60 FPS. This greatly reduced idle GPU usage, which means now it is possible to have even a 100 `fluid.jit.plotter` instances open and sitting around without a noticable overhead. Because of the "resize watcher" script, drawing will still tick at 1 FPS on idle (so not 100% on-demand), but that did not seem to cost a lot in my tests, so I left it in there. In my experience the change also made the interaction look smoother on a high refresh rate screen.

*(not sure why this PR still has commits from the previous PR, I must have messed something up, but if I can fix it, let me know, the new commits are only the ones from April this year)*


balintlaczko:
I'm actually thinking of a valid `setpoint` implementation now, just to make it even more aligned with `fluid.plotter`, but not sure about the exact details (or performance cost) yet. If you wanna hold on with merging, I can come back to you in a bit about whether it worked or not.

balintlaczko:
`setpoint` is almost done, only a few edge cases to iron out. But in the meantime I noticed a quite grave limitation of `jit.pworld` (for this use case) that it is always rendering at 60 fps by default. Which means just having several (idle & empty) `fluid.jit.plotter`-s will generate a significant GPU overhead. I'm testing the switch to `jit.gl.render` instead and make every rendered frame on-demand. Hopefully this will mean a huge efficiency boost.

tremblap:
Don't worry - this will go in 1.0.8 which is not in the very short term - there is a max threading experiment to do before with @AlexHarker and Rod (all for Max) so i will need to spend more time in Max...

balintlaczko:
Just to keep you guys posted, I changed the rendering to on-demand, now I can have open a 100 `fluid.jit.plotter`s without any overhead.

Another HUGE news, is that I figured out how to do individual point sizes with the `jit.gl.mesh` via a custom shader, which means, I will be able to do 100% parity with the original `fluid.plotter`, finally! All the functionality, total interface match, but everything much faster. Which means painfree drop-in replacement. Yaaaay. Some nice commits coming soon! :)

balintlaczko:
Okay dear people, the PR is ready for review. Please, try it and try to break it, beacuse implementing these new stuff (especially `setpoint`) added a lot of complexity and new "code". On my stuff it seems to behave well. Still have to test on Windows, though, will report here when that's done.

balintlaczko:
"One more thing": I added a workaround so that the `jit.gl.mesh` point sizes do not change on mac os on a non-retina external screen. This is an already reported Max bug (if @benbrackenc74 is watching), which is that Max does not communicate to the render shader of `jit.gl.mesh` whether the screen is a retina screen or not. Mac os will report a screen resolution that is half the width and half the height to Max. With the new fix I went into the lengthy hack to check:
- which screens are *probably* retina (via `jit.displays`: assuming that if the current resolution is 1/4th of the max resolution, we are on retina)
- monitoring the location of the patcher that hosts `fluid.jit.plotter` and deducing which screen that must be
- if we are on a mac (via `gestalt`) --> if the host patcher is on a "non-retina" screen --> apply a 0.5 scaling on the point sizes of the `jit.gl.mesh`es. 

This fixes the problem on my setup, I can move the patcher window (containing a `fluid.jit.plotter`) to the external screen and back, and the point sizes update to look the same.

N.B: this only works if the patch was first opened (or `fluid.jit.plotter` was created) on the retina screen, and _then_ we move it to the external screen. Otherwise when moving from the external (non-retina) screen to the retina screen, we can observe another bug with `jit.gl.mesh` (also mentioned in the same bug report): that the view "zooms" to the lower left quadrant -- this is unfixable, you have to delete & undo the `fluid.jit.plotter` on the retina screen.

tremblap:
hmmmm and what is the behaviour between 2 types of screen without the fix? I'm super worried about the one-way-a-way-another-way-the-other-way behaviour...

tremblap:
Hello

I reckon this is ready to review now? One point before I do so:

> But if you don't want this to be part of the abstraction I get that and I can take it out, or maybe not have it on by default, just let me know what you think.

If it takes any resource at all, let's make it at least switchable? or is the resource so low that we can ignore?

balintlaczko:
> If it takes any resource at all, let's make it at least switchable? or is the resource so low that we can ignore?

I would say it consumes very little resources, but probably not nothing. Here is how it works:
- upon object creation we create a dictionary of available screens, their coords and whether or not they are retina (assuming that if the resolution (count of pixels) if 1/4th of the reported screen resolution). This happens only once, and I haven't noticed any lag because of it.
- then, in the js script that we use to monitor the size and position of the bpatcher in edit/presentation mode (to position the `jit.pwindow` acordingly) we have one more line to also report the parent patcher window's location (`outlet(1, this.patcher.parentpatcher.wind.location);`). This is passed into a subpatch where it is immediately gated out if the platform is not mac. 
- If it is mac, then based on the window coords we identify which screen that is, and from the dictionary we query whether it is retina or not.
- Whenever this flag changes we apply an additional scaling to the points and refresh, once.

It sounds like a lot when I spell it out like that but I really haven't noticed any overhead. The periodic query in the js script also slows down to a 1000ms rate if it detects that nothing changes.

tremblap:
this sounds sensible. if Rod has not complained, I will test and merge like this.

balintlaczko:
Okay! Let me also test everything on Windows too.

balintlaczko:
Everything seems to be OK on Windows too!

balintlaczko:
Hold on with the merging, new update incoming (that fixes the retina stuff and does less ugly things when resizing).

jamesb93:
Did you want me to re-review this @balintlaczko ?

balintlaczko:
I think it was @tremblap who did that, but if you could hold on a bit, @rconstanzo found some potential issues that are still waiting for me to investigate. But almost there :)

tremblap:
indeed I am in standby

rconstanzo:
Just found some funky edge behaviors with `setpoint` and related messages, the way different color points "overlap" (this happens differently in both versions (perhaps unavoidably?)), as well as the `highlight -1` message (in `fluid.plotter` it clears stuff but throws an error in `fluid.jit.plotter`).

tremblap:
@balintlaczko you'll let me know when you think it is ready to deep check?

balintlaczko:
Yeah, I still have some testing/fixing to do, but I'll report here when all is done!

jamesb93:
👀

tremblap:
@balintlaczko if we were to release soon, is your revision almost there or we skip a version?

balintlaczko:
Hey! Sorry, I'm still delaying the additional improvements/issues highlighted by @rconstanzo, so if you have to release very soon (in a week), then I'd say go ahead, and I'll hop on next time. Things got quite busy at work, but I'm onto the plotter fun as soon as possible, but I'll still need around 2 weeks (depending on the severity of the problems I find).

balintlaczko:
Okay, did some fixes and improvements now based on @rconstanzo's earlier feedback.

Changelog:
----------

- "highlight -1" hides the highlighted mesh. Parity in behavior with fluid.plotter.

- Do not reset xrange/yrange upon new dataset. Parity in behavior with fluid.plotter.

- Fix setpoint with 0 default point size.

- Enable blending, handle alpha color values and disable depth for all non-highlighted points. This can improve visibility for dense areas.

- Set default point color to "0 0 0 0.8" for parity with fluid.plotter.

- Automatically fill an alpha value of 1. if user only specifies RGB in "setpoint".

- Fix non-theme-aware bubble comments in fluid.plotter.maxhelp and fluid.jit.plotter.maxhelp.

tremblap:
thanks for this beautiful, I will aim to look at it in the next days (but I'm on the road so it might take a week or so)

tremblap:
(although in the meantime there are conflicts to be solved - don't forget to merge in the tip of the main branch and solve them @balintlaczko please, I can help it needs be)

balintlaczko:
Thanks, yes, I'll figure that out. :)

jamesb93:
> Thanks, yes, I'll figure that out. :)

If you need any help with a tricky rebase – let me know and I can help get this one ready for merge.

balintlaczko:
Phew. I'd like to thank @jamesb93 for his MASSIVE help with this. I ended up doing the manual json surgery merging the following way.
1. Cloned the main branch on one laptop, while having my PR on another,
2. Opening the two affected help files (*fluid.plotter.maxhelp* and *fluid.jit.plotter.maxhelp*) and carefully checked that I carried over all changes while keeping mine. These were all just comment box text color changes from fixed to theme-aware ones. For *fluid.jit.plotter.maxpat* I assumed that my PR version wins. (Only these 3 files were in conflict.)
3. When all changes from the main were also commited to this PR branch (again, just comment text color changes) I created local backups of all 3 PR version files, then I went into the online merge resolve thing, and I *thought* that I chose my PR version for every case (which I assumed was *under* the `=======`). I commited this and [merged](https://github.com/flucoma/flucoma-max/pull/417/commits/8f816d9300f00b9d6984bebf31e29585fe964264).
4. Then pulled & checked, only to realize that the old changes were kept instead of my PR version (which also had now all the real changes from the old main at this point). I probably should have chosen the bits *above* the `=======`? (It was a bit confusing since both were labelled "main"...)
5. Nevermind, now I restored all 3 files from my local backup and pushed it again (the commit ["reverse-merge from local bkp"](https://github.com/flucoma/flucoma-max/pull/417/commits/7de6c0f069fb969b2dc17e3b64c483ae3ac107e7)).
6. Wrote this log and felt ashamed.

Double checked that the current state of things are correct. Hopefully ready for a review. :)

balintlaczko:
Just a teaser for improving morale, compare this to the same plot from the earlier version (scroll to the top see that one):
image

Now it actually adheres to the alpha values in point colors, so when there are a ton of overlapping points, this could help you see through the mess (a bit). Here all point alphas are 0.2.

jamesb93:
> Just a teaser for improving morale, compare this to the same plot from the earlier version (scroll to the top see that one): image
> 
> Now it actually adheres to the alpha values in point colors, so when there are a ton of overlapping points, this could help you see through the mess (a bit). Here all point alphas are 0.2.

Love this!

tremblap:
ok this is amazing

tremblap:
@jamesb93 and @balintlaczko I've corrected a typo and pushed. can I merge now?

tremblap:
(btw @balintlaczko it seems my push has not notify here, so there is something strange with this PR) Let me know if you didn't get 966a7c4)

balintlaczko:
Hmm, I don't get the changes on my end. What was the change visually? Maybe I can just implement it and push it here? Otherwise merging is fine by me!

balintlaczko:
I get this banner:
image

tremblap:
fun. The only change was that in Max8, the tab 'advanced' the `shape square` and `shape circle` are too low so the cable is ugly :)

balintlaczko:
Ah then the huge diff could have been pushing from Max 8, I guess. I tried to fix it now, check and let me know!

tremblap:
perfect. let's merge.

jamesb93:
> perfect. let's merge as soon as @jamesb93 says yes

I say yes!!!

PR #428, opened 2024-10-08
Add sidebar logo support

tremblap:
this addition will help with supporting the sidebar logo for the package- there is no conflict with 7 and 8


jamesb93 (approved):
fab

jamesb93:
I would hold fire before merging, just for me to check that the dimensions of the SVG can be what they are now, or if they should be square.

tremblap:
how could we test that? We could add a fake snippet in the package to see if it appears?

PR #435, opened 2024-11-08
remove dependency caches and bootpaths

jamesb93:
These refer to machine specific paths that shouldn't exist in production.

We can safely remove them without affecting how the patches work.


weefuzzy (approved):

jamesb93:
> wow was there an automatic action to remove those? I can see that I might have saved the work on the patches and push those without noticing...

There is a preference in Max that can cause it. I just removed it manually though.

image

tremblap:
I reckon you search the whole package for it with your regexp ninja skills?

jamesb93:
> I reckon you search the whole package for it with your regexp ninja skills?

Absolutely not ninja but yes... :)

tremblap (approved):

PR #447, opened 2025-05-24
dataset:tobuffer: adds a flag to flip the ID-label

tremblap:
this is linked to https://github.com/flucoma/flucoma-core/pull/305

### DIFF
--- a/help/fluid.dataset~.maxhelp
+++ b/help/fluid.dataset~.maxhelp
@@ -57,7 +57,7 @@
 						}
 ,
 						"classnamespace" : "box",
-						"rect" : [ 48.0, 208.0, 854.0, 698.0 ],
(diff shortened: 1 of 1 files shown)


tremblap:
@balintlaczko feel free to add a small example - at the moment there is only a 'from dataset' in the tab, so we could make a sub-patch, or another tab. It needs to be terse, let's not go waxlyrical

balintlaczko:
On it!

tremblap:
> On it!

do you see when I mention you in the PR? @weefuzzy has a much wiser proposal as usual...

balintlaczko:
Hey yes, thanks for the nudge! I'll cook up something today, and we can refine it from there. :)

PR #456, opened 2025-11-09
Jit.plotter: quick fix of broken mouse modifier parsing

balintlaczko:
It is really just one commit: 7414d8d67446cf13f2f76a6a9848f76ef8cee8ae

But for some reason it shows the entire history of the jit.plotter here. If it is a problem, please help me figure out the fix. :)

In that one commit I fixed the mousing (cycling introduced additional numbers reported at the end of the mouse list, and that broke my earlier logic), and applied the ;max sortpatcherdictonsave 1.

So this is just meant to be a quick and small PR, I'll put the new features in following ones.


tremblap:
@jamesb93 is a git ninja he will I hope share his tricks :)

balintlaczko:
if it is too complex, I can just delete this fork, fork again, and redo that one commit? or is that too ugly in other ways? just tell me what is the "cleanest" sequence of making a new PR, because I have some other ones planned, and instead of having one enormous PR, I'd rather do a series of small PRs this time...

jamesb93:
The issue is that the base of this branch (where it was branched from) is very old. I suspect its base is the old `fluid.jit.plotter` PR. When we merged that PR, the 92 commits were squashed and rewritten, and so the 92 commits from that PR are now seen as "new" commits in this one.

Two ways to solve it:

`git rebase` this branch from main and drop the 92 preceding commits and `pick` only the meaningful one.

To do that you can do it entirely on the CLI by going...

`git rebase -i main`

then an editor will spawn and look like so

image

change the word `pick` to `drop` for every commit you want to exclude, which is every single commit except the last (bottom) one.

Then save that text file and exit and the rebase will work. Then:

`git push --force-with-lease`

---

create a new branch whose base is the updated state of `main`, and cherry pick the meaningful change over from that branch into this one. Then make a new PR.

In general I'd advise the first method, if only for learning how it can be useful for completely rewriting history when necessary.

balintlaczko:
Upon saying `git rebase -i main` my terminal said this:
image
Somehow no entries to pick from...

I sense pain further down this road. Since I'd have to drop everything but one commit, maybe it is better if I start from scratch with a clean slate. I can then redo the fix (it is really not complicated).

Issue #463, opened 2026-05-20
Feature Request: Labelset should return something else than an empty string if getlabel is called with a non-existent label


balintlaczko:
### Is your feature request related to a problem? Please describe the problem.

Currently, when I use `getlabel` with a non-existent identifier, I get an empty string (symbol) as a response, which is not very convenient in Max.

### Describe the solution you'd like to see.

In Max, it is atypical to have an empty atom (symbol/string/list). It would be more convenient to just not get anything in return (just the "getlabel"). So, currently, we get a list: ["getlabel", ""]. If it was just ["getlabel"], it would already be much more convenient to handle. But I'm open to other solutions too.

### Describe alternatives you've considered

My current hack:

<img width="530" height="707" alt="Image" src="https://github.com/user-attachments/assets/7a0f8303-9be2-4af4-a68c-35deb40b969a" />

It works, but it would be nicer to get something easier to handle.

### Additional context

This came up in the `setpoint` implementation for `fluid.jit.plotter`. Since the message can imply a new or existing identifier, I have to check if it exists in the dataset/labelset. I guess a separate `contains` message/method could also work, but it wouldn't be necessary if we could avoid outputting the empty symbol.

tremblap:
I had a discusson about a 'isPoing' method in the (slow) pipeline with @weefuzzy - we could do 'isLabel' or use 'contains' indeed, but he will have (very well informed) opinions...

tremblap:
His previous wisdom was: "IAC, if one were to add a message to the containers, I think it should be called `contains` and you'd probably want to call std::unorderd_map::count to see if a key exists"

tremblap:
flucoma/flucoma-core#344 is in progress.

PR #464, opened 2026-05-20
fluid.jit.plotter data loading & color update

balintlaczko:
Changelog:
----------

- changed default background color to 0.95 0.95 0.95 0.95 1 for parity with fluid.plotter.

- revamped dataset loading now uses tobuffer labelflip to avoid needing a for-loop to get the ids-to-samples mapping. (Unfortunately, now we need two tobuffer calls to build ids-to-samples and samples-to-ids, but that is still much faster than the previous for-loop.)

- added option to load 1 to 7-dimensional datasets for the combined loading of positions, sizes, and/or colors

- added dimmap message: remap input dimensions (think of zl.indexmap)

- added lut and lutoffset messages: choose from different color maps (see: colorcet.com)

- added misc/CET-colormaps.jxf: 60 color maps as a binarized jitter matrix

- help patch/advanced tab: fixed dataset not clearing

- help patch: added new tab "more dimensions & color maps"

- updated reference

image


balintlaczko:
Note that I removed the misc folder from gitignore, because that's where I intended to put the new CET-colormaps.jxf file. Let me know if you want me to add a file in another way. I also think you may need to add some script to copy it into the PM-ready distribution? (@jamesb93 what do you think?)

balintlaczko:
...just some descriptive info: basically, it is a lot faster now, because it avoids a for-loop that was necessary to perform at least once to map samples to IDs (now uses the new labelflip option in the labelset). It also lets you refer to a >2D dataset with all the position, size, and color information, so you don't have to set those with loops. And, thanks to @rconstanzo's suggestions, now there is also a 1D color option and [59 new perceptually uniform color maps](https://colorcet.com/gallery.html) and an HSL on the 60th slot.

rconstanzo:
Having played with this for a couple weeks, I have to say this is super useful addition.

Not only super fast/performant, but being able to avoid using `pointcolor` to set the color/size across a huge `fluid.dataset~`'s worth of data is huge.

tremblap:
Thanks for this, it looks awesome! 

It'll take time to check, because we need to change the cmake (we don't change the gitignore in a PR!)

I'll spend quality time over it in the next weeks, hopefully towards the next version

balintlaczko:
Haha, I knew you would like it. 😆

Just let me know how you prefer it, and I'll deliver it that way instead. 
No rush.

jamesb93:
@balintlaczko if you take a look here: https://github.com/flucoma/flucoma-max/blob/d685140748a277c6da5c3588fa005735031f7aa2/CMakeLists.txt#L183 you can see how things end up in misc at build time. However, the data is pulled through from: https://github.com/flucoma/flucoma-core/tree/main/Resources/Data. 

Is the colour scheme file you've created something that is specific to Max, or could it feasibly be used in other environments? If the answer is yes, it is specific to Max, is the `misc` folder the right place for it? If the answer is yes again, then maybe we should unignore `misc` and merge the addition of the `Data` files from core and any local files in misc at build time.

balintlaczko:
Specific to Max? Yes, because it is an exported jitter matrix binary format (.jxf). (And it is only used by the `fluid.jit.plotter` which is also Max-only.)

Misc folder? I figured since all the other "data" files (such as datasets) are also here, this may be the most suitable folder.

Does that mean that I should not change anything (for now)?

jamesb93:
Indeed, I actually think this is fine :) Have you tested the packaging with `cmake` to ensure the package which gets made contains the `misc` folder?

tremblap:
The misc folder is problematic to include for local builds during dev code iirc. Let me check.

jamesb93:
> The misc folder is problematic to include for local builds during dev code iirc. Let me check.

I don't quite yet understand how, as it's ignored and up to the dev to make it functional. I do this by copying the `json` I need from the core repo. We could setup the ignore so that it ignores everything except the file we choose to as well, which would preserve this workflow.

balintlaczko:
> Have you tested the packaging with cmake to ensure the package which gets made contains the misc folder?

No, never tried, but I will after there is an agreement about the gitignore status. :)

> We could setup the ignore so that it ignores everything except the file we choose to as well, which would preserve this workflow.

Sounds like a good compromise to me.

I'll just wait for @tremblap to check things, I'm ready to make any changes.

tremblap:
@balintlaczko I'm still puzzled by the CET as Matrix. What is the gain there, and could it be a gain for the other CCEs if it was in a format like a normal dictionary as json file (which could be imported by the others too in whatever format they need)

the other question is how the 7 dims is generally useful, vs idiosyncratic. For instance, CataRT used x-y-colour-shape. I'm mostly interested to discuss all of this in term of data visualisation and use, vs classification viz (which is what we currently do)

Generally, this PR is for me 4 PRs, here they are re-bundled:

### fixes:
- changed default background color
- revamped dataset loading now uses tobuffer labelflip
- help patch/advanced tab: fixed dataset not clearing

### new feature 1: new colour schemes
- added lut and lutoffset messages: choose from different color maps (see: colorcet.com)
- added misc/CET-colormaps.jxf: 60 color maps as a binarized jitter matrix

### new feature 2: added dimmap message

### new feature 3: more dimensions


And apart from fixes (which is a quick merge as uncontroversial) the others need each their own discussion on interface decisions and ways forward. Does that make sense?

rconstanzo:
There's nothing lost by adding the dimensions, but I have to say having used >2d plotting for the last few years, it is definitely useful. (I know you're not a believer of >2d!)

I personally consider it a 4d space (x, y, size, color), but I treat color as a 1d space (via the LUTs). @balintlaczko just exploses the RGBA as 4 separate dimensions so you can set your own color in a more granular way, if you don't want the baked in LUTs.

I do agree that 7 "real" dimensions are likely not useful.

///////////////////////////

re: breaking things up across multiple features.

I don't see how the color stuff would be separate from more dimensions as they only really apply to >2d datasets anyways (in the current implementation). There's something to be said for having the LUTs be generally available across the CCEs, and I suspect the compression was just considering storage/download space in the package. 60 LUTs is not a huge amount of space though.

balintlaczko:
Hey, thanks for the feedback, @tremblap & @rconstanzo!

Why have the LUTs in a .jxf format?
1. Since this was a Max-specific feature (as—sadly, as of now—the whole of `fluid.jit.plotter`), I thought it did not make sense to package this in a CCE-agnostic .json (since the other .js files & shaders the plotter needed were also kept Max-specific). But: if you prefer the .json, I am happy to convert it, it will then just add a slight hanging when you create the first `fluid.jit.plotter` object, since then I have to just-in-time read into dataset > tobuffer > to matrix. (From the 2nd instance onwards it's fine.)
2. Since I thought this was Max-only, I tried to find the format that consumes the least amount of space. Having this many numbers in .json or .txt will be 4-5x larger in my estimate (but still probably under 1MB). If that is fine, I of course can add it as a .json (in a fluid.dataset format).

Why 7D:
Personally, in my research, I only needed so far to get direct access to RGBA, since I always had my custom functions to encode "something" as color. @rconstanzo enlightened me about the handiness of LUTs, so that you can treat the color domain as a direct mapping target (e.g., after dimension reduction), and he pointed me to Kovesi's research too about the perceptually uniform color maps. I thought this is a great idea, so I implemented it. IMO I think it is useful to have both; in my other dev work, I usually have the philosophy to include both the high-level-and-easy mode, and the let-me-do-it-myself mode (pytorch-lightning has been one of my models), and I see the LUTs and RGBA correspond to those modes, respectively.

Multiple PRs:
You are totally right, I hyperfocused (again) and one thing led to another... :) It would be _slightly_ uncomfortable now though to disentangle the features into separate PRs (because of object positioning, and other visual patching issues), but if you say we should do it like that, I am happy to retire this one and go step by step. Alternatively, we could discuss the choices here, and tweak this PR to establish a consensus.
Just one more thing to this: note that I don't support setting the LUTx point-by-point, because it is a lot more efficient to load all at once as part of the dataset. We could debate this, but my hot take is that the individual `setpoint`, `pointsize`, and `pointcolor` messages also promote an inefficient workflow, and if it was not a matter of parity, I would retire them.

So, suggest me changes, or a preferred roadmap, and I move on from there! :)

tremblap:
It all depends how quick you need the fix. There are 3 threads needing discussing, plus the fix, so that will take time and iterations. It can get messy here if we keep all entangled, but if you are on no rush, then we can leave it here (but that is not my favourite option but I know that Max makes PRs difficult because of bad diffs).

In the meantime, this is progressing: flucoma/flucoma-core#344

you will notice the ugly problem (again for max) 'containsId' becoming 'containsid' which made me think of the film Ice Age. but feel free to compile to see if it sorts your life r.e. flucoma/flucoma-max#463

rconstanzo:
I guess I'm biased here, but I don't know that the scope of the changes really extend past `fluid.jit.plotter`, so these all seem like features of that one object, and as long as `fluid.jit.plotter` has feature parity with `fluid.plotter`, that's what's most important.

Whether or not the LUTs get propagated to other CCEs (or `fluid.plotter`) or whether other features end up elsewhere too (e.g `dimmap`) would determine if/when more broad interface discussions would be relevant.

For what it's worth, I do think those features would be cool elsewhere, and I very much agree with @balintlaczko 's approach of having easy and complex implementations of features, so being able to broadly set colors as a dimension in `fluid.plotter` is something I always wanted (`dk.plotter` is basically a wrapper around `fluid.plotter` to make it do all of that).

And as a small technical side-note, my request for having the 4d version (where you set the colors via LUTs and 1d data) even though @balintlaczko had already built a 7d version was that even though I could set the colors however I wanted with a 7d dataset, it meant that you couldn't easily change out LUTs or offset them without `for loop`-ing it all in again, which becomes super slow at the scale of points where `fluid.jit.plotter` is useful.

jamesb93:
my 2c would be that if we wanted to broaden the use of these new colour additions then it would involve significant work putting them in the other CCEs (testing, checking, someone to actually do all of this work for free and for the benefit of the wider community). IMV, we should not hinder the work of anyone who has the gumption to draw up meaningful improvements too much on the premise of universally designing any and all features, especially given how localised this is to the world of `fluid.jit.plotter` which is already divergent.

my only criticism/idea (which may stem from ignorance and naivety) is that it could be better to instead design a "CET loading interface" and then provide one through the core repository that can be loaded. That way future work doesn't become about aligning `fluid.jit.plotter` and everything else, but just adding a CET loader, if that makes sense. Format becomes more fluid then too, if its JSON, size doesn't really alarm me in anyway and again makes it more pliable for future work when and if it needs to be added to other environments.

written in haste so hopefully making sense :--)

tremblap:
I certainly agree with @jamesb93 's 2nd comment. as for parity, we can discuss more. removing CET because it can be loaded in RGBA is the way forward to lower dims. We still need to discuss that thoroughly in relation to potential shape too, and sizes, so all of this might require multiple messages. This is a good lab for interface.

as for 'promoting inefficient workflow' this is true from a computer scientist perspective. but from a musician starting to think in batches, thinking of them as many items is a good way in.

and we have not discussed dimmap, which again seem to bundle in one object what can be done in others. but I might be missing the point.

at the moment, colours, and sizes, and shapes, are more geared towards classification in the basic visualisation. thinking about making them 'continuous' dimensions, for shape as well as for colour, is quite arbitrary and controversial already, so let's use this PR to think clearly about it all... I know we all have opinions here, so let's try to think widely in range of uses and users and proficiencies.

rconstanzo:
Not sure I follow your comment here. Are you saying removing the CET stuff from `fluid.jit.plotter` functionality completely because you can `setpoint`? Or leaving the 7d (RGBA approach) as the way to achieve 4d (xy, size, color) functionality?

Shapes are nice, but that seems like a whole other idea/PR. You can set shapes now but AFAIK it applies to all objects, so it's more a "style" thing to change the shapes rather than anything functional.

The `dimmap` idea emerged from me suggesting the ability to "rotate" dimensions (e.g. swapping size and color) just to get a different view on the same data. This is the post on the FluCoMa forum that I used as a point of reference:
[https://discourse.flucoma.org/t/fluid-datasetplot-abstraction-for-2d-3d-visualization/783/48](https://discourse.flucoma.org/t/fluid-datasetplot-abstraction-for-2d-3d-visualization/783/48)

This can be done with `fluid.datasetquery~` already (or other approaches), but it was more a QoL suggestion to match up with the offset message for the LUTs so you could "rotate" the LUT to get a different view of the data, as well as "rotate" the dimensions for a different view. Even though what's happening is technically very different, conceptually I view those as the same kind of thing.

tremblap:
that is the problem with multiple PRs folded in one :)

let's leave the fixes (pr#1) we have 3 intertwined subjects, all of which sharing or not solutions

- the desire for batch colour assignation
- the desire for more continuous values as dimensions (and the desire to rotate these)

Tension underlying flucoma interface design, as ever: 
- supporting learning for entry users, as well as power moves by more advanced users
- finding general solutions that are worth the maintenance time
- parity across CCEs to foster discussions beyond ghettos

hence the careful moves. so for me for now, I am trying to widen the scope of the 3 PRs, then see what is idiosyncratic (growing zooming out) to you two, to Maxers, to advanced creative coders, and to data-manipulators in CCEs.

So far, I'm still unclear about the questions above, hence asking and proposing and challenging and trying to find the most elegant way forward.

rconstanzo:
I guess an easy solution would be to de-jitterfy the CET stuff (or perhaps have a parallel `.json` version alongside the jitter matrix version that would be specific to `fluid.jit.plotter` to avoid the load-time issues that @balintlaczko mentions. Or perhaps having the `.json` version live in the non-Max version of the repo but the `.mjx` one inside the Max repo (unless there's appetite or interesting in expanding the color functionality to `fluid.plotter` too.

I guess in a more zoomed out way, I do think having a LUT vs RBGA interface enables sufficient surface area for both ends of the user spectrum, and perhaps incorporating @jamesb93 's suggestion of a way to be able to load/extend the LUT interface could further broaden out the entry user-side of things, as well as keeping things a bit more future-proof-y where more LUTs can be added or incorporated without needing to jump ship and roll your own via the RGBA interface if you still want to manipulate color with only a single dimension.

Things like `dimmap` do have broader implications, and as flexible `fluid.datasetquery~` is, I absolutely *hate* the interface of that object. I find it to be paralyzingly confusing to the point that most of the time, I would prefer to dump things out to a `dict` (or `dict` -> `coll`) and then manipulate the data there, rather than having to work out whatever awkward messages I need to give to `fluid.datasetquery~`.

That is perhaps a bit too zoomed out in the scope of this PR, but that is to say that "manipulating datasets" (or even worse, labelsets...) is currently not a pleasant experience and because it was added near the end of the project, IMO, it didn't get the TLC in terms of interface that other things did.

balintlaczko:
Hey guys, thanks for the discussion! Sorry, the end of the term got in the way, but I'm back on this horse now. I see now that implementing 42534 features in one PR is not a good idea. :) I am now trying to find my way through the maze, to see what we can do the get this rolling again.
From the original changelog:
> changed default background color to 0.95 0.95 0.95 0.95 1 for parity with fluid.plotter.

I guess this is fine.

> revamped dataset loading now uses tobuffer labelflip to avoid needing a for-loop to get the ids-to-samples mapping.

Probably this one too.

> help patch/advanced tab: fixed dataset not clearing

Should be fine.

> added option to load 1 to 7-dimensional datasets for the combined loading of positions, sizes, and/or colors
> added misc/CET-colormaps.jxf: 60 color maps as a binarized jitter matrix
> added lut and lutoffset messages: choose from different color maps (see: colorcet.com)
> added dimmap message: remap input dimensions (think of zl.indexmap)
> help patch: added new tab "more dimensions & color maps"
> updated reference

Here is where we start to get into the weeds. The problem is that these features are quite major changes, so it's more a design question rather than just an optimization one. And they are entangled. For what it's worth, I have been using it in my research a lot lately in 4D, and I definitely think it is a huge improvement in two ways:
1. being able to visualize continuous data dims as size and color (from a LUT) _does_ make sense, especially when the data is not noise, but there is some correlation between these dimensions. In my perception, this often turns projections (from PCA or UMAP) "pseudo-3D", as if we'd look at a perspective image of a 3D point cloud. Roughly the same goes for color maps, especially if using the linear ones (L4 is my go-to).
2. Loading multi-dimensional datasets mapping extra dimensions to size/color is so much more efficient than setting size/color point-by-point. It also lets the user save everything as one dataset and load it later conveniently.

My attempt to disentangle this would be to have the multi-dim dataset loading as an optimization as one feature, the colormap stuff as another, and finally dimmap (to be dropped).

We currently have `pointsize` (1D) and `pointcolor` (4D) for point-by-point assignments. To get these loaded with the XY coordinates, we could have the following loading options. If the dataset is:
- 2D: interpret it as XY (currently there)
- 3D: interpret it as XY + pointsize
- 5D: interpret it as XY + pointcolor (where alpha=1.0)
- 6D: interpret it as XY + pointcolor
- 7D: interpret it as XY + pointsize + pointcolor.

This is something that could be implemented in all plotters in all CCEs. That could be one PR / CCE.

Now LUTs. To make this cross-platform, I agree, there needs to be a .json-format colormap (probably only the CET colors, popping the HSL? so we can clearly attribute it to CET & Kovesi et al.). Then, every plotter in all CCEs has the responsibility of handling/parsing this .json in a suitable way. I am not sure a "unified" loader is necessary, since I think the different plotters probably need to handle it differently (unless we bring all of them into C++, for which I don't have the time). The unification IMO _is_ the .json format.

This could be another PR / CCE + one global _CET_colormaps.json_. This PR should also update the plotters to accept 4D datasets (as XY + pointsize + LUT x), the `lut` method to choose a map, and the `lutoffset` to offset it (especially useful for rainbow and circular type colormaps).

Finally `dimmap`: I think this maybe too case-specific, we could drop this and say that it is a user responsibility to remap dims with a datasetquery? Since the multi-dim loading greatly speeds things up, it's not a big loss in performance anyway.

#### For this PR:
- I assume I could keep all the fixes and optimizations
- Could also implement the multi-dim loading for the `jsui`-based `fluid.plotter`
- Provide the color maps as .json (or perhaps this should go to flucoma-core?) and handle it individually in `fluid.plotter` and `fluid.jit.plotter`
- Undo `dimmap`

#### Then the Pd PR:
- add multi-dim loading, LUT-handling to the Pd fluid.plotter

#### Then the SC PR:
- same there for the SC plotter

What do you think of this plan @tremblap and the rest? :)

tremblap:
ok if you promise it is the last megagigamulticontroversialfeaturepr you do, then let's continue this discussion here.

- removing dimmap is a good idea
- colours as json on core is a good idea
- which brings us to the magic 2 to 7 mapping. the only case that is not covered is alternative 6d where someone wants size and colour(without alpha). I know @weefuzzy has always been allergic to magic orders, but at the same time we had no choice for a few (for instance, order of selected spectral shapes)

So I'd say, go for it, and in the meantime @weefuzzy might pop in and propose an idea for a less magic interface. or we go with it and we see how it floats.

balintlaczko:
megagigamulticontroversialfeaturepr is my middle name, haha. But I can also break it up into several ones, or also open about demagification suggestions. :)

